PR #15, opened 2024-03-12
Proposal 2: `did:fractal`

brianorwhatever:
This approach is a bit different and out there but it attempts to shift the turtles to make them easier to reason about. The `scid` is derived from a rotation of a well known DID document, which allows it to effectively be created before the DID document is constructed.

I didn't quite hammer out all the details and get everything from my brain into this PR but it should be good enough to have the conversation of whether we are interested in the approach.


brianorwhatever:
Ideas from this have been worked into #18

Issue #3, opened 2024-03-07
How to authenticate a DID rotation


brianorwhatever:
The method must only allow rotations of DID documents by the `controller`.

Controller is a bit of a loaded term but my assumption is that this is the `controller` property of the DID doc. If so, the first document will be lacking this property won't it? What do we expect a controller property to look like? It can be a set of keys but we would have to further specify (see #2)

swcurran:
After thinking through this and texting with Drummond and @peacekeeper, I think this is the best we can do if we want to enforce that only an authorized key (or keys) can update the DIDDoc of a DID. The first part of this covers the single key use case of a `verificationMethod`, and lower down in the comment I propose adding in multi-sig via `verifiableCondition`.

* We want to have a proof on a DIDDoc from a key (or keys) authorized to update the DIDDoc.
* Per the DID spec, the top level `controller` item (string or set) contains the DID(s) of the party(ies) authorized to update the DIDDoc — e.g. to add a new version to the DIDDoc
    * However, by the spec, the controller must be a DID without a fragment, meaning it can’t reference a verificationMethod in the DIDDoc, so it is not a straightforward reference to a specific key within the DIDDoc.
* Thus:
    * The top level `controller` is the DID or DIDs authorized to update that DIDDoc.
        * If the controller references one or more different DID(s), resolve those DID(s), else the DID is self-referencing.
        * If the controller item is missing, assume the DID is self-referencing.
    * Each verificationMethod in the controller DID(s) is authorized to sign/update this DID's DIDDoc.
        * An issue I see with that is a given `verificationMethod` might be present to be used for purposes other than to update the DIDDoc, such as to sign a VC. But so be it. By inclusion, it is authorized.
        * A way around that is to have DIDs that control updates to DIDs used for other purposes. So an entity might have one controller DID for updating a set issuer DIDs that can’t update themselves, but rather are only updated by the one controller DID.

For multi-sig, we define that a `verifiableCondition` is used equivalently to a `verificationMethod` for the purposes of updating the DIDDoc of a DID:

* Use [verifiableCondition2021](https://www.w3.org/TR/did-spec-registries/#verifiablecondition2021) for multi-signature management.
* A verifiableCondition can be used wherever a verificationMethod can be used.
* Notably in this DIDMethod, as proof of authorization to update a DIDDoc.
    * That is, the DIDDoc of a controller DID contains a verifiableCondition entry.
[...]

PR #17, opened 2024-03-19
Proposal #3 - `did:log`

brianorwhatever:
Uses JSON Patch to create a list of JSON diffs from the current doc -> new doc


brianorwhatever:
Ideas from this have been worked into #18

Issue #4, opened 2024-03-07
Document Format


brianorwhatever:
Proposals:

## Proposal 1: something very DID docy

Benefits:
- Easy parsing at resolution time
Drawbacks:
- Large file size

```json
{
  "@context": [
    "https://www.w3.org/ns/did/v1",
    "https://w3id.org/security/multikey/v1"
  ],
  "authentication": ["#z6Mko4t2y5Pbx96hbQP5p8JkJXA3z16nAq7VpUyZGcjhd9ra"],
  "capabilityInvocation": ["#z6MkqHVcPwdxaNnwAYEi1ZrqE1z3EonVdZ6hmGFXz9wSHqUB"],
  "capabilityDelegation": ["#z6Mkhn3U7DCskv9SzLW8Fwa3ssYhwjzZMTPj443RqMmaBKYb"],
  "assertionMethod": ["#z6MkmUNA3PrAr28xhofDKMN6So2tjsC2HoohsJzMwejXuxAm"],
  "verificationMethod": [{
      "id": "#z6Mko4t2y5Pbx96hbQP5p8JkJXA3z16nAq7VpUyZGcjhd9ra",
      "type": "Multikey",
      "publicKeyMultibase": "z6Mko4t2y5Pbx96hbQP5p8JkJXA3z16nAq7VpUyZGcjhd9ra"
    }, {
      "id": "#z6MkqHVcPwdxaNnwAYEi1ZrqE1z3EonVdZ6hmGFXz9wSHqUB",
      "type": "Multikey",
      "publicKeyMultibase": "z6MkqHVcPwdxaNnwAYEi1ZrqE1z3EonVdZ6hmGFXz9wSHqUB"
    }, {
      "id": "#z6Mkhn3U7DCskv9SzLW8Fwa3ssYhwjzZMTPj443RqMmaBKYb",
      "type": "Multikey",
      "publicKeyMultibase": "z6Mkhn3U7DCskv9SzLW8Fwa3ssYhwjzZMTPj443RqMmaBKYb"
    }, {
      "id": "#z6MkmUNA3PrAr28xhofDKMN6So2tjsC2HoohsJzMwejXuxAm",
      "type": "Multikey",
      "publicKeyMultibase": "z6MkmUNA3PrAr28xhofDKMN6So2tjsC2HoohsJzMwejXuxAm"
    }],
  "service": [{
    "id": "#didcomm-1",
    "type": "DIDCommMessaging",
    "serviceEndpoint": {
[...]

PR #18, opened 2024-03-25
Proposal #4 `did:tdw` -- TDW = TrustDID Web

brianorwhatever:
This version moves things into the log format we've been discussing. Still need to do the signatures on updates and check a few things on resolution.


in/keys.json
```diff
@@ -0,0 +1 @@
...3H2Xq3KfaGB7Lz3NkDXyLDapcATXviWr7kM8nSPQBYKjJjMPYKDJkz3d1iJZqBy"},{"publicKeyMultibase":"z6MkiJ1ZcwLgaFiMf45x9uV79CXz9nBBsU9pZsHi6Si2NFu3","secretKeyMultibase":"zrv4yokKuCScoHft73M7AWcCEXEo61toVz3UBr998Nf1tU6qZKpbyop41Gqyf85rzDMx6cMYZMXvZXTkexcmvZtFih7"},{"publicKeyMultibase":"z6Mkpzr75VJEbHsjVt1bAi7ianzAWUhMBnH6MyWsMgrxSh5Y","secretKeyMultibase":"zrv4yokKuCScoHft73M7AWcCEXEo61toVz3UBr998Nf1tU6qZKpbyop41Gqyf85rzDMx6cMYZMXvZXTkexcmvZtFih7"},{"publicKeyMultibase":"z6MkvD5C46SMRubbNsqekgrPLZn5dRoWw2VyHHNMGhDKqRhh","secretKeyMultibase":"zrv22eZkn4dcQfZQNFpa9BoUpZgSnA7w9czxipz1FLu8QhXALYRTnZByPJ2f6tZ19dwgDqRT8DRGkGEN3HDV8WL9wdh"},{"publicKeyMultibase":"z6MkuaP6TUBMMN6EXrbpFcvdg6SWB6eYZLEVSwZhq9FLtLgQ","secretKeyMultibase":"zrv22eZkn4dcQfZQNFpa9BoUpZgSnA7w9czxipz1FLu8QhXALYRTnZByPJ2f6tZ19dwgDqRT8DRGkGEN3HDV8WL9wdh"},{"publicKeyMultibase":"z6MkkEMjh9oGHk7sjVLXj4LGZrpaCtQZqVChpcyXwb64Yncr","secretKeyMultibase":"zrv1Y1ZQk5t1V29kUS1h5LVNBKfhLRZQDxXBytxKVqr1tVuvr3qNuw16DXFVCoZ6Kw3SmqDPk7khmzACJZGJYYRDXbp"},{"publicKeyMultibase":"z6MkgGYhu9CWRvXQQpN2dprWn6wgeR5c7TcF19FsyLWJVRo1","secretKeyMultibase":"zrv1Y1ZQk5t1V29kUS1h5LVNBKfhLRZQDxXBytxKVqr1tVuvr3qNuw16DXFVCoZ6Kw3SmqDPk7khmzACJZGJYYRDXbp"}]
```
swcurran, line 1: OK … I’m curious. Why is there two entries for every secret key, each with a different public key?  Not important — I’m just wondering.

out/mxK7ZzHRWaF3seULxR6T5Vym/log.4.txt
```diff
@@ -0,0 +1,5 @@
+["history:1","did:tdw:1","mxK7ZzHRWaF3seULxR6T5Vym",{}]
```
swcurran, line 1: Interesting…the log is 3474 bytes, and gzip compresses to 1005. Pretty small going across the wire...

brianorwhatever: yeah, I am missing proofs in all of the updates which are kinda big but yeah pretty promising

in/keys.json
brianorwhatever: I don't think there is? formatted a bit better
```
[
   {
      "publicKeyMultibase":"z6MkpM2PEcqPZgGW5RUc3bB7iNUd9XD9BgoSRu8y8QJqnPFF",
      "secretKeyMultibase":"zrv4TX25xCGYTKg17dPHL4zbLLWmFoSdDeoxR9mkgoBwkASLbWFcz4VXVVc3mhdaKMWne8QmEzZehA3mGNPFBpeTRCV"
   },
   {
      "publicKeyMultibase":"z6MksudZk7NtBD9Tn2kMCAX68LmQCWiTsVnWJrUPmQiQDj7U",
      "secretKeyMultibase":"zrv4TX25xCGYTKg17dPHL4zbLLWmFoSdDeoxR9mkgoBwkASLbWFcz4VXVVc3mhdaKMWne8QmEzZehA3mGNPFBpeTRCV"
   },
   {
      "publicKeyMultibase":"z6MkfBTm5X9oMESxDs8y91W6N9NrY9fa8FKTihJ5VpAq3fGy",
      "secretKeyMultibase":"zrv5BRv3jbJkKaiBCfN4xw8FTJbH2XY58Th8VtJ9KkHaBw3UsA5B3yQPgZDQZ3H6jDmQEPhjbKuqg6A8waBt6KjCZkH"
   },
   {
      "publicKeyMultibase":"z6MkmMbsLn2tEfzETDo6kAxPLM27Q2w271HKakfz4ty4UBsW",
      "secretKeyMultibase":"zrv5BRv3jbJkKaiBCfN4xw8FTJbH2XY58Th8VtJ9KkHaBw3UsA5B3yQPgZDQZ3H6jDmQEPhjbKuqg6A8waBt6KjCZkH"
   },
   {
      "publicKeyMultibase":"z6MkugKEvvj4KqKhwuuKn8kJG64Lv5aSbZd79R26iU1rw2dx",
      "secretKeyMultibase":"zrv5UqLSxHP1oReqww6Can4495WPGSwpMmV7EvhWW4R3X3FkfcfFnhBUs4e7g795cgXTgdsDnWCPkPCyktS3EgXzAok"
   },
   {
      "publicKeyMultibase":"z6MktwC1PPHWmRsEfwkFbkycFipygux5P57bTVvrdBxNfgrQ",
      "secretKeyMultibase":"zrv5UqLSxHP1oReqww6Can4495WPGSwpMmV7EvhWW4R3X3FkfcfFnhBUs4e7g795cgXTgdsDnWCPkPCyktS3EgXzAok"
   }
...
]
```

brianorwhatever: oh.. err I see it now.. Hmm must be something broken with when I generated them. Definitely not intentional

brianorwhatever: Fixed the `keys.json` file to include proper keypairs

out/mxK7ZzHRWaF3seULxR6T5Vym/log.4.txt
brianorwhatever: All the proofs are in now and verified

swcurran: Compression is good — about the same. gz’d file is about 30% of the full file.

src/method.ts
```diff
@@ -0,0 +1,285 @@
...
+export const deriveHash = async (input: any): Promise<{logEntryHash: string}> => {
+  const data = canonicalize(input);
+  const hash = await sha256.digest(Buffer.from(data));
+  return {logEntryHash: base58btc.encode(hash.digest)};
+}
+
+export const createDID = async (options: CreateDIDInterface): Promise<{did: string, doc: any, meta: any, log: DIDLog}> => {
+  let {doc} = await createDIDDoc({...options, controller: `did:${METHOD}:${PLACEHOLDER}`});
+  const {logEntryHash: genesisDocHash} = await deriveHash(doc);
+  const {scid} = await createSCID(genesisDocHash);
```
brianorwhatever, line 40: I am generating a log entry hash for the "0th" item that isn't in the log which is just the same hash algo except instead of using the log entry I am using the genesis doc. I am then deriving the `scid` from this hash (grabbing the last 24 characters). This hash is used as the "previous hash" for version 1 instead of the empty string.

When resolving this the process is reversed and the placeholder is put back in and then the previous hash is generated to be used as the input to verify the first line in the log.

src/interfaces.d.ts
```diff
@@ -0,0 +1,72 @@
...
+  logFormat: string,
+  baseProtocol: string,
+  scid: string,
+  extensions: any
+];
+type DIDLogEntry = [
+  logEntryHash: string,
+  versionId: number,
+  timestamp: string,
+  params: {method?: string, scid?: string},
```
brianorwhatever, line 30: @swcurran @andrewwhitehead I did find it **much** easier to include the scid as a `param` on the first log line.

test/logs/test-run.log
```diff
@@ -0,0 +1,213 @@
...
+    {
+      id: "did:tdw:jNQ4MKr87HBoox9iXKjRKt3S#4jDJpwPh",
+      controller: "did:tdw:jNQ4MKr87HBoox9iXKjRKt3S",
+      type: "Multikey",
+      publicKeyMultibase: "z6MkkmrDWT9n8rmAVfEvuyBFroc6RFNffAoycrLw4jDJpwPh",
+    }, {
+      id: "did:tdw:jNQ4MKr87HBoox9iXKjRKt3S#DuAWQAnf",
+      controller: "did:key:z6MkhoA3GQu8Sek1GFrMpmZNFxy2fSXVzzMggdndDuAWQAnf",
+      type: "Multikey",
+      publicKeyMultibase: "z6MkhoA3GQu8Sek1GFrMpmZNFxy2fSXVzzMggdndDuAWQAnf",
```
brianorwhatever, line 173: @swcurran @andrewwhitehead adding an external controller like this ^

brianorwhatever:
I originally had included the list of static keys in keys.json as Bun was messing up the generate key function and always returning the same key no matter what. Looks like that is fixed somewhere in between bun versions 1.0.26 and the current 1.0.35. Thought I would have to look into it later but it looks like I don't.. Still leaving the keys in there though as it's nice to use a static list for the tests. Makes things nice and deterministic.

brianorwhatever:
Moved to https://github.com/bcgov/trustdidweb-ts/pull/1

PR #24, opened 2024-04-05
Initial version of the spec

swcurran:
Lots more to do, but I wanted to get something published.


brianorwhatever (approved):
Haven't had a chance to go through this but we can get it onto main and iterate from there

PR #27, opened 2024-04-08
Add DID URL and /whois resolution sections

swcurran:
DID URL resolution -- implicit and with overrides.


spec/specification.md
```diff
@@ -403,88 +406,101 @@ When processing a [[ref: DID log entry]] where the `pre-rotation` parameter is a
...
+publish about itself. The intention is that anyone interested in a particular
+`did:tdw` DID can resolve the a `<did>/whois` DID URL, and if returned, if it
+contains any useful (to the resolver) Verifiable Credentials that might help in
+learning more about who is the controller of the DID and if they should be
+trusted. It is up to the [[ref: DID Controller]] to decide to publish a `whois.vp`
+verifiable presentation and if so, which verifiable credentials to put into the
+verifiable presentation.
+
+`did:tdw` DIDs **automatically** supports a `#whois` service endpoint with the
+following definition based on the [[xref:LINKED-VP,Linked VP Specification]], with the `serviceEndpoint`
```
brianorwhatever, line 436: This will try to find a term `Linked VP Specification` in the linked vp spec

spec/specification.md
```diff
@@ -403,88 +406,101 @@ When processing a [[ref: DID log entry]] where the `pre-rotation` parameter is a
...
+[[xref:VCDM,VCDM Credential]]s) the [[ref: DID Controller]] has decided to
+publish about itself. The intention is that anyone interested in a particular
+`did:tdw` DID can resolve the a `<did>/whois` DID URL, and if returned, if it
+contains any useful (to the resolver) Verifiable Credentials that might help in
+learning more about who is the controller of the DID and if they should be
+trusted. It is up to the [[ref: DID Controller]] to decide to publish a `whois.vp`
+verifiable presentation and if so, which verifiable credentials to put into the
+verifiable presentation.
+
+`did:tdw` DIDs **automatically** supports a `#whois` service endpoint with the
```
brianorwhatever, line 435: Is the reason to have this automatic just to force people into using it? I worry this complicates resolver code for no gain since if you are putting a file somewhere you want to be discoverable updating/creating a service seems like a reasonable process to expect to have to go through

brianorwhatever (commented):
I don't think we should create new file extensions (`.vp` or `.vc`) and should just use `.json`. It would open up too many other considerations outside of what we're trying to do here.

specs.json
```diff
@@ -33,7 +33,13 @@
           "module": true
         }
       ],
-      "katex": true
+      "katex": true,
+      "external_specs": [
+        {
+          "LINKED-VP": "https://identity.foundation/linked-vp/",
+          "VCDM": "https://www.w3.org/TR/vc-data-model/"
```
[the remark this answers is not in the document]
swcurran: Yes — and xref doesn’t do in the spec what I wanted. I’ll convert to “regular” links.

spec/specification.md
swcurran: Yeah — I realized that after.  Removing.

spec/specification.md
swcurran: The reason is to make it dead simple to use and sell people on the idea that every DID method should support “/whois”.  It is such a great idea — and would make discoverablity so much easier, particularly for the “do I trust the issuer of this VC?” question.

brianorwhatever (approved):

PR #29, opened 2024-04-08
change to 'authorized keys'

brianorwhatever:


swcurran:
@brianorwhatever — you have to update to the base branch.

swcurran (approved):

swcurran:
@brianorwhatever -- I think we should be .gitignore'ing the `index.html` file. Do you agree?  The GHA generates it, right?

brianorwhatever:
Yep should be definitely

PR #32, opened 2024-04-09
With .gitignore change the index is missing -- trying update

swcurran:
@brianorwhatever -- the index.html on gh-pages is now missing.

Took a guess at a fix, but please look.  Does the `require` you have do a render of the page?  I thought so because of the "nowatch" parameter, but since there is no index.html, I explicitly added an `npm run render`.  Please take a look ASAP as we have shared the URL a few places.

Just looked at the output of the `npm run render` command and see that it is the same as the command following...

### DIFF
--- a/.github/workflows/render-specs.yml
+++ b/.github/workflows/render-specs.yml
@@ -20,7 +20,8 @@ jobs:
       - name: Install and Build 🔧 # This example project is built using npm and outputs the result to the 'build' folder. Replace with the commands required to build your project, or remove this step entirely if your site is pre-built.
         run: |
           npm install
-          node -e "require('spec-up')({ nowatch: true })"
+          npm run render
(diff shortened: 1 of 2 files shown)


brianorwhatever:
The issue is that .gitignore is being used by the deploy. I've created #33 which removes it before deploy which should work

.github/workflows/render-specs.yml
```diff
@@ -20,7 +20,8 @@ jobs:
       - name: Install and Build 🔧 # This example project is built using npm and outputs the result to the 'build' folder. Replace with the commands required to build your project, or remove this step entirely if your site is pre-built.
         run: |
           npm install
-          node -e "require('spec-up')({ nowatch: true })"
+          npm run render
```
brianorwhatever, line 23:
```suggestion
```

PR #34, opened 2024-04-09
Adjustments to the spec -- especially the JSON for the implicit services

swcurran:
Updates to the spec. I had to merge some changes I had made with your @brianorwhatever . I think got it right, but take a look.  I really want to get the `service` JSON corrected, so a quick review would be appreciated.


brianorwhatever (approved):
Looks good!

PR #48, opened 2024-04-15
Minor adjustment to tdw-did production

andrewwhitehead:


spec/specification.md
```diff
@@ -43,7 +43,7 @@ segment in the (optional) URL path. See examples below. The `domain-segment` and
 would inevitably be wrong.
 
 ```abnf
-tdw-did = "did:tdw:" *( domain-segment “.” ) scid 1*( “.” domain-segment ) ( “.” domain-segment ) *( ":" path-segment )
+tdw-did = "did:tdw:" *( domain-segment “.” ) scid 2*( “.” domain-segment ) *( ":" path-segment )
```
swcurran, line 46: It is actually correct with the 1.  Drummond did these and we went through them. Not the extra domain segment after the “1*” one. So it is say “0 or more sub-domains + SCID + one or more sub-domains and domain + the final TLD”

So the 2 would not be right.

swcurran (changes requested):
Well — really just cancel the PR. :-)

spec/specification.md
andrewwhitehead: I replaced `1*( “.” domain-segment ) ( “.” domain-segment )` with `2*( “.” domain-segment )` which is equivalent as far as I can tell

swcurran: No because that would you mean you have to have 3 segments after the SCID, and you only need two.

SCID.example.com is all that is required.

andrewwhitehead: Not sure where you're getting 3 from?

swcurran: `scid 1*( “.” domain-segment ) ( “.” domain-segment )`

That’s SCID + 1 or more DS + DS — that’s 2 or more.

If we put the 2 where you want it, we get SCID + 2 or more DS + DS, which is 3 or more.

andrewwhitehead: I'm not just replacing a 1 with a 2 though, it's replacing the extra domain segment

swcurran:
OK - :-).  You are right — I missed that.  I knew when I started this, I would not “win" this discussion. And it is easier to understand with what you suggest.

swcurran (approved):

PR #52, opened 2024-04-29
chore: updating dependabot file to support gha, TS and JS packages

rajpalc7:


rajpalc7:
Hi @WadeBarnes , this is ready for review now.

PR #57, opened 2024-06-03
Change how the authorized keys are published and used

swcurran:
I've updated the spec. to include the changes we discussed last week with regards to authorized keys, SCID generation, and pre-rotation. While I generally followed what we discussed, I updated the SCID generation in a way I think makes the most sense, varying from what we discussed. Please review and let me know what you think.

`updateKeys` is a list of `did:key` DIDs in the parameters, as we discussed.

The existing SCID generation was based on just the DIDDoc with placeholders. I changed that to make it the Log Entry (minus the Data Integrity proof, of course), but including the parameters, the versionId, versionTime and `"value": "<DIDDoc>"` with placeholders (literal string `{SCID}`) wherever it is needed. It seemed easier to understand to use the entire entry vs. just a subset.

@andrewwhitehead - I also didn't do the SAID approach of using a `########...` placeholder because I thought it would be a pain for the DIDDoc writer. It is much easier to just use the `{SCID}` placeholder, and to explain it, and really doesn't make any difference.

Pre-rotation becomes MUCH easier than what we had before. No more "detecting" key changes in the DIDDoc, and the hashing is just of the `did:key` DIDs, not the DIDDoc key reference entry data structure.

I really struggled over the description of what are the active authorized keys for the initial entry and subsequent entries. I'll probably try ChatGPT to see if I can make it better. Suggestions gratefully accepted...

### DIFF
--- a/spec/abstract.md
+++ b/spec/abstract.md
@@ -1,12 +1,13 @@
 ## Abstract
 
-The `did:tdw` (Trust DID Web) method is an enhancement to the
-`did:web` protocol, providing a complementary web-based DID method that addresses limitations
-of `did:web`. Its features include:
+The `did:tdw` (Trust DID Web) is an enhancement to the `did:web` DID method,
+providing a complementary web-based features that address `did:web`'s
+limitations. `did:tdw` features include:
 
 - Ongoing publishing of all DID Document (DIDDoc) versions for a DID instead of,
-  or alongside a `did:web` DID/DIDDoc.
+  or alongside a current `did:web` DID/DIDDoc.
 - The same DID-to-HTTPS transformation as `did:web`.
+- Supports the same [High Assurance DID-to-DNS] mechanism.
 - The ability to resolve the full history of the DID using a verifiable chain of
   updates to the DIDDoc from genesis to deactivation.
 - A [[def: self-certifying identifier]] (SCID) for the DID that is globally
@@ -14,7 +15,7 @@ of `did:web`. Its features include:
   enables [[ref: DID portability]], such as moving the DID's web location (and
   so changing the DID string itself) while retaining a connection to the
   predecessor DID(s) and the DID's verifiable history.
-- DIDDoc updates include a proof signed by the DID Controller(s) *authorized* to
+- DIDDoc updates contain a proof signed by the controller(s) *authorized* to
   update the DID.
 - An optional mechanism for publishing "pre-rotation" keys to prevent the loss of
   control of a DID in cases where an active private key is compromised.
@@ -25,6 +26,8 @@ of `did:web`. Its features include:
   published by the DID controller) a [[ref: Verifiable Presentation]] containing
   [[ref: Verifiable Credentials]] with the DID as the `credentialSubject`,
   signed by the DID.
+
+[DID-to-DNS]: https://datatracker.ietf.org/doc/draft-carter-high-assurance-dids-with-dns/
 
 Combined, the additional features enable greater trust and security without
 compromising the simplicity of `did:web`. The incorporation of the DID Core
--- a/spec/abstract.md
+++ b/spec/abstract.md
@@ -1,7 +1,7 @@
 ## Abstract
 
-The `did:tdw` (Trust DID Web) is an enhancement to the `did:web` DID method,
-providing a complementary web-based features that address `did:web`'s
+Trust DID Web (`did:tdw`) is an enhancement to the `did:web` DID method,
+providing complementary web-based features that address `did:web`'s
 limitations. `did:tdw` features include:
 
 - Ongoing publishing of all DID Document (DIDDoc) versions for a DID instead of,
--- a/spec/specification.md
+++ b/spec/specification.md
@@ -109,9 +109,14 @@ Creating a `did:tdw` DID is done by carrying out the following steps.
    defines the permitted [[ref: parameters]]. The authorization [[spec DID KEY]] DIDs
    generated in step 3 **MUST** be included in the [[ref: parameters]] of the
    first version of the DID, in the `updateKeys` item.
-6. Pass the DID string, initial DIDDoc, and [[ref: parameters]] to a `did:tdw`
+6. Pass the initial DIDDoc, and [[ref: parameters]] to a `did:tdw`
    "Create" implementation that **MUST**:
-   1. Calculate the [[ref: SCID]] for the DID as defined in the [SCID Generation
+   1. Extract from the DIDDoc the value of the `id` item, the DID identifier
+      itself, and verify that it is a valid `did:tdw` DID that contains the
+      `{SCID}` placeholder in a allowed location in the DID string as per the
+      ABNF of a `did:tdw` DID as defined in the [Method-Specific
+      Identifier](#method-specific-identifier) section of this specification.
+   2. Calculate the [[ref: SCID]] for the DID as defined in the [SCID Generation
       and Validation](#scid-generation-and-validation) section of this
       specification.
    2. Replace in the DIDDoc the placeholder for the [[ref: SCID]] `{SCID}` with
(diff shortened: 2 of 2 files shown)


spec/specification.md
```diff
@@ -93,46 +93,48 @@ Creating a `did:tdw` DID is done by carrying out the following steps.
...
-      The definition of "authorized" in this case is specified in the
-      [Authorized Keys](#authorized-keys) section of this specification. The
-      proof becomes the sixth and last JSON item in the DID log entry.
+      authorized key from a [[spec DID KEY]] DID in the `updateKeys` item in the [[ref: parameters]], and the `entryHash` as the proof `challenge`.
+      The proof becomes the sixth and last JSON item in the DID log entry.
    7. Put the resulting entry, with extraneous white space removed as the
-      contents of a file `did.jsonl` and publish the file at the appropriate
-      location defined by the `did:tdw` value.
+      contents of a file `did.jsonl`.
+	 8. Publish the file at the appropriate location defined by the `did:tdw` DID identifier.
```
brianorwhatever, line 137: I think the indentation got messed up here

spec/specification.md
```diff
@@ -200,18 +202,21 @@ the following steps:
...
+   9. If any verifications fail, discard the DID as invalid.
+   10. As each log entry is processed and verified, collect the following information
       about each version:
-      1. DIDDocument
-      2. `versionId`
-      3. `versionTime`
+      1. DIDDoc.
+      2. `versionId` of the DIDDoc.
+      3. `versionTime`of the DIDDoc.
+			4. The active [[spec DID KEY]] DIDs authorized to update the DID, from the `updateKeys` lists in the [[ref: parameters]].
```
brianorwhatever, line 216: Indentation issue?

spec/specification.md
```diff
@@ -375,14 +379,12 @@ items are defined below.
...
-- `nextKeys`: An array of strings that are hashes of future keys to be added to
-  the DIDDoc.
+- `nextKeys`: An array of strings that are hashes of [[spec DID KEY]] DIDs that will be added to the `updateKeys` list in a future version of the DID.
   - The process for generating the hashes is defined in the [Pre-Rotation Key
     Hash Generation and
     Validation](#pre-rotation-key-hash-generation-and-validation) section of
     this specification.
-  - If the parameter `prerotation` has been set to `true`, all keys added to a
-    version of the DIDDoc after version 1 **MUST** have a corresponding hash
+  - If the parameter `prerotation` has been set to `true`, all [[spec DID KEY]] DIDs added to `updateKeys` lists **MUST** have a corresponding hash
```
brianorwhatever, line 387:
```suggestion
  - If the parameter `prerotation` has been set to `true`, all [[spec DID KEY]] DIDs added to `updateKeys` lists **MUST** have a corresponding hash in the `nextKeys` list
```

spec/specification.md
```diff
@@ -413,23 +415,26 @@ items are defined below.
...
 
 Where:
 
-1. The `initial DIDDoc with placeholders` is the initial DID Doc defined by the
-   [[ref: DID Controller]] with the placeholder `{SCID}` put everywhere the
-   [[ref: SCID]] will be used in the resolved version 1 DIDDoc. At minimum, the
-   `{SCID}` **MUST** appear in the top level `id` item of the DIDDoc. It **MAY**
-   occur elsewhere in the DIDDoc.
+1. The `partial initial log entry with placeholders` consists of the following elements of the
+   the first log entry, as a JSON Lines array. The placeholder is the literal string "`{SCID}`".
```
brianorwhatever, line 430:
```suggestion
   first log entry, as a JSON Lines array. The placeholder is the literal string "`{SCID}`".
```

spec/specification.md
```diff
@@ -449,15 +454,13 @@ To verify the [[ref: SCID]] of a `did:tdw` DID being resolved, the resolver
...
    1. If less than 28 characters, terminate the resolution process with an
       error.
-3. Extract from the first [[ref: DID log entry]] the `value` item's value, which
-   is the initial DIDDoc.
-4. Treat the `value`s value as a string and do a text replacement of the `scid`
-   from the first step with `{SCID}`. This should result in the `initial DIDDoc
-   with placeholders` data needed for the next step.
+3. Remove from the first [[ref: DID log entry]] the data integrity proof.
+4. Replace the `entryHash` value with the placeholder literal "`{SCID}`"
+4. Treat the rsulting log entry as a string and do a text replacement of the `scid`
```
brianorwhatever, line 459:
```suggestion
4. Treat the resulting log entry as a string and do a text replacement of the `scid`
```

spec/abstract.md
```diff
@@ -1,20 +1,21 @@
 ## Abstract
 
-The `did:tdw` (Trust DID Web) method is an enhancement to the
-`did:web` protocol, providing a complementary web-based DID method that addresses limitations
-of `did:web`. Its features include:
+The `did:tdw` (Trust DID Web) is an enhancement to the `did:web` DID method,
```
brianorwhatever, line 3:
```suggestion
Trust DID Web (`did:tdw`) is an enhancement to the `did:web` DID method,
```

spec/abstract.md
```diff
@@ -1,20 +1,21 @@
...
+providing a complementary web-based features that address `did:web`'s
```
brianorwhatever, line 4:
```suggestion
providing complementary web-based features that address `did:web`'s
```

spec/specification.md
```diff
@@ -93,46 +93,48 @@ Creating a `did:tdw` DID is done by carrying out the following steps.
...
+   that is not required.
+5. Define a JSON array of valid [[ref: parameters]] that affect the generation
    of the DID. The [DID Generation and Validation
    Parameters](#didtdw-did-method-parameters) section of this specification
-   defines the permitted [[ref: parameters]].
-4. Pass the DID string, initial DIDDoc, and [[ref: parameters]] to a `did:tdw`
+   defines the permitted [[ref: parameters]]. The authorization [[spec DID KEY]] DIDs
+   generated in step 3 **MUST** be included in the [[ref: parameters]] of the
+   first version of the DID, in the `updateKeys` item.
+6. Pass the DID string, initial DIDDoc, and [[ref: parameters]] to a `did:tdw`
```
brianorwhatever, line 112: What is the "DID string" here? I don't think the DID will be known yet

brianorwhatever (approved):

swcurran:
@brianorwhatever -- The comments you made were mostly tweaks to the wording. Does that mean you are more or less happy with the approach? Especially around the SCID generation?

brianorwhatever:
I only briefly went through the scid generation and was pretty tired. I'm hopefully going to look at implementing it today so let's hold off on merging until after that

swcurran:
Great. Definitely!

brianorwhatever:
@swcurran I believe I implemented what is here in https://github.com/bcgov/trustdidweb-ts/pull/9 and it seemed to make sense. Take a look there and see if you agree I matched what's written here. Seemed like the SCID generation alg made sense. Interested in hearing @andrewwhitehead feedback

swcurran:
Middle of the night thinking. I think the nextKeys array should not collect used/unused ones as I proposed. Rather, each instance becomes the current one, and the previously current set of hashes are dropped. I think a bad implementation could allow any nextKey to be reused, which would be a vulnerability. 

The more general rule for (I think) any parameter is that the most recent value is the current one.

Reasonable?

brianorwhatever:
I had just reached the same thought. Makes implementation easier as well 👌

swcurran:
Good stuff. I'll update that. Makes the spec easier as well.

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: I think this was fine — the “missing text” you identified is in the next line.

spec/specification.md
swcurran: You’re right — not really needed and removed for now. It has to be the `id` of the DID, but the “create” process does not have to know it.

I think the create function **MUST** extract the `id` from the DIDDoc, verify it is a `did:tdw` and that it has the SCID placeholder in it.

Added that check.

brianorwhatever:
I think to differentiate we should use

`updateKeys` and `nextKeyHashes`

swcurran:
> I think to differentiate we should use
> 
> `updateKeys` and `nextKeyHashes`

Done.

brianorwhatever (approved):

PR #62, opened 2024-07-01
Update to spec after DICE, adjustments and adding approvers concept

swcurran:
Updates to the spec per discussions with @brianorwhatever and @andrewwhitehead following the DICE event. Adds the "approver" concept to the DID Method (aka witnesses) for feedback.  Note that I went with the suggestion from Brian on how to implement approver proofs, as I think they are the cleanest.

Still to do -- an implementers guide section on approvers, and a massive cleanup of the "refs" and "defs".  Much to do there...

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ of `did:web`. It's features include the following.
   update the DID.
 - An optional mechanism for publishing "pre-rotation" keys to prevent loss of
   control of the DID in cases where an active private key is compromised.
+- An optional mechanism for having collaborating "approvers" (aka "witnesses")
+  that approve of updates to the DID by the DID Controller before publication.
 - DID URL path handling that defaults (but can be overridden) to automatically
   resolving `<did>/path/to/file` by using a comparable DID-to-HTTPS translation
   as for the DIDDoc.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -28,6 +28,8 @@ of `did:web`. It's features include the following.
   update the DID.
 - An optional mechanism for publishing "pre-rotation" keys to prevent loss of
   control of the DID in cases where an active private key is compromised.
+- An optional mechanism for having collaborating "approvers" (aka "witnesses")
```
brianorwhatever, line 31: I'm not sure I like the "approvers" term. I know you have a history with the word "endorser" but it makes more sense here. Approvers feels like you are losing your self-sovereignty.. won't die on this hill though

brianorwhatever (dismissed):

README.md
swcurran: Should we just go with witnesses?  I think approvers is not bad, since to be valid, they need approval.

swcurran: Updated throughout to witness/witnesses. The term "approve" is left in in place vs. the verb "witness", but that is for readability -- not having the word "witness" in a paragraph 6 times :-).

brianorwhatever (approved):

Issue #76, opened 2024-07-18
Versioning the specification


bj-ms:
It would be helpful to version the draft version of the specification to make it easier for implementers to know which version of did:tdw they support.

swcurran:
Addressed via #77

Issue #74, opened 2024-07-17
Keeping the SCID independent of the HTTPS URL


andrewwhitehead:
Using ONLY the proposed syntax `did:tdw:{SCID}:domain:path`, and not allowing for the SCID to occur in the HTTPS URL simplifies the spec significantly and would be beneficial for security in removing any ambiguity.

If the SCID is allowed to be case-sensitive (which it cannot be in a subdomain), then we could also use base58 for the encoding and potentially reduce the minimum length of the SCID at the same security level for second preimage attacks.

While `did:tdw:{SCID}.domain.example` or `did:tdw:domain.example:{SCID}` looks nice and readable, it would also allow for `did:tdw:{SCID1}.domain.example:{SCID2}` or `did:tdw:domain.example:{SCID1}:{SCID2}` meaning that either SCID could be the one defined by the log file.

swcurran:
@brianorwhatever , @dmitrizagidulin — do you agree? 

My leaning, having written and explained the spec, is that this is the right thing to do. A SCID as a subdomain is kinda silly and unlikely.  A SCID as file path is a bit more reasonable, but there are lots of things that can be done here.

If we do switch to base58 we would eliminate the option of putting in the SCID twice in the DID and so that it appears once in the HTTPS URL, e.g. `did:tdw:{SCID}:example.com:dids:{SCID}`

So my thought is to go with what is proposed.

andrewwhitehead:
Base58 for a subdomain or path component (or query parameter or whatever) is actually fine, it only becomes a problem when we rely on it for security. If the DID was `did:tdw:{SCID}.domain.example` then any different capitalization of the SCID would be considered equivalent because domain names are case insensitive. You could end up resolving a different subdomain, with a different log file and a slightly different SCID (although achieving that hash collision would still be difficult, you would have many more options).

If the SCID is enforced separately from the URL then I don't think there is an issue, although we might still want to recommend against using it as a subdomain to avoid ambiguity. The DID Core spec does say that case sensitivity within the method-specific ID is up to the DID method.

swcurran:
To summarize, the changes to be made are:

- SCID is required in the first element of the method-specific identifier `did:tdw:<scid>:<did web method specific identifier>`
- If the SCID is only in the required location in the DID, it will not be in the HTTPS address.
- SCID MAY be placed in the method-specific identifier in the domain or path elements of the `<did web method specific identifier>`
  - In that case, it will appear multiple times in the DID
[...]

PR #77, opened 2024-07-19
SCID in fixed location and add version with changelog

swcurran:
As per #74, places the SCID into a fixed location in the DID, resulting in DID-to-HTTPS transformation that matches did:web. This removes the SCID from the HTTPS URL, so added a comment on creating noting that the SCID can be optionally added into the domain or path components of the DID without impact on the DID or processing.  In making this change, I did not change the SCID generation algorithm -- although we could do that (changing to case-sensitive base58 encoding, and shortening the SCID to 24 characters). Still debating if we should make that change.  Please add your thoughts!

Added the Specification Version to the top of the spec (per #76 ) with a link to a Version Changelog at end of the specification.

Tightened up the language around the creation of the DID -- changing the previous "verify that..." to "**MUST**" where applicable.

### DIFF
--- a/spec/specification.md
+++ b/spec/specification.md
@@ -18,16 +18,12 @@ specified below.
 
 The `did:tdw` method-specific identifier contains both the [[ref:
 self-certifying identifier]] (SCID) for the DID, and a fully qualified domain
-name that is secured by a TLS/SSL certificate, with an optional path. Given the
+(with an optional path) name that is secured by a TLS/SSL certificate. Given the
 DID, a [transformation to an HTTPS URL](#the-did-to-https-transformation) can be
 performed such that the [[ref: DID Log]] for the `did:tdw` DID is retrieved (via
 an `HTTP GET`) and processed to produce the DIDDoc for the DID. As per the
-Augmented Backus-Naur Form (ABNF) notation below, the SCID **MUST** be a part of
-the method-specific identifier by one of the following mechanisms:
-
-- A separate, colon-delimited element at the start of the method-specific identifier.
-- Embedded in the fully qualified domain name.
-- Embedded in the optional path relative to the fully qualified domain name.
+Augmented Backus-Naur Form (ABNF) notation below, the SCID **MUST** be the first element of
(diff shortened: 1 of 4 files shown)


dmitrizagidulin (approved):
👍 
Nice (also good job keeping a changelog doc)

spec/specification.md
```diff
@@ -18,16 +18,12 @@ specified below.
 
 The `did:tdw` method-specific identifier contains both the [[ref:
 self-certifying identifier]] (SCID) for the DID, and a fully qualified domain
-name that is secured by a TLS/SSL certificate, with an optional path. Given the
+(with an optional path) name that is secured by a TLS/SSL certificate. Given the
```
andrewwhitehead, line 21:
```suggestion
name (with an optional path) that is secured by a TLS/SSL certificate. Given the
```

swcurran: Fixed.

spec/specification.md
```diff
@@ -48,21 +44,20 @@ examples below. The `domain-segment` and `path-segment` elements refer to
 here the full ABNF of those elements from that RFC would inevitably be wrong.
 
 ```abnf
-tdw-did = "did:tdw:" scid ":" *( “.” domain-segment ) *( ":" path-segment )
-tdw-did = "did:tdw:" *( domain-segment “.” ) scid 2*( “.” domain-segment ) *( ":" path-segment )
-tdw-did = "did:tdw:" 2*( domain-segment “.” ) *( ":" path-segment ) (":" scid ) *( ":" path-segment )
+tdw-did = "did:tdw:" scid ":" 2+( “.” domain-segment ) *( ":" path-segment )
```
andrewwhitehead, line 47:
```suggestion
tdw-did = "did:tdw:" scid ":" domain-segment *( "." domain-segment ) *( ":" path-segment )
```
I think your editor added some smart quotes too

swcurran: Thanks — fixed.

andrewwhitehead (approved):

PR #79, opened 2024-07-23
Use Base58 instead of Base32 and Remove SCID shortening

bj-ms:
As part of the discussion and security concerns in https://github.com/bcgov/trustdidweb/issues/75, I propose to change Base32 to Base 58 and remove the shortening of the SCID during it's generation.

I also changed the version, changelog  and added myself as an author (feel free to comment on a possible rollback on that part).

### DIFF
--- a/spec/definitions.md
+++ b/spec/definitions.md
@@ -78,12 +78,12 @@ when forced to change (such as when an organization is acquired by another,
 resulting in a change of domain names) and when changing DID hosting service
 providers.
 
-[[def: base32_lower]]
+[[def: base58]]
 
-~ Applies [[spec:rfc4648]] to convert
-data to a `base32` encoding, and then lower cases the result. Data encoded as
-base32 consists of a string of characters containing only the letters A-Z and
-digits 2-7.
+~ Applies [[spec:draft-msporny-base58-03]] to convert
+data to a `base58` encoding. Data encoded as
+base58 consists of a string of characters containing only the letters A-Z, a-z and
+digits 1-9.
 
 [[def: Linked-VP, Linked Verifiable Presentation]]
 
--- a/spec/definitions.md
+++ b/spec/definitions.md
@@ -84,9 +84,7 @@ providers.
 [[def: base58btc]]
 
 ~ Applies [[spec:draft-msporny-base58-03]] to convert
-data to a `base58` encoding. Data encoded as
-base58 consists of a string of characters containing only the letters A-Z, a-z and
-digits 1-9.
+data to a `base58` encoding.
 
 [[def: Linked-VP, Linked Verifiable Presentation]]
 
--- a/spec/definitions.md
+++ b/spec/definitions.md
@@ -81,7 +81,7 @@ when forced to change (such as when an organization is acquired by another,
 resulting in a change of domain names) and when changing DID hosting service
 providers.
 
-[[def: base58]]
+[[def: base58btc]]
 
 ~ Applies [[spec:draft-msporny-base58-03]] to convert
 data to a `base58` encoding. Data encoded as
--- a/spec/definitions.md
+++ b/spec/definitions.md
@@ -42,6 +42,9 @@ make changes to a DID document.
 the data of a log entry. The generated entry hash is subsequently put into the
 log entry and *MUST* be verified by a resolver.
 
+[[def: multihash]]
+~ Per the [[spec:draft-multiformats-multihash-07]], Multihash is a specification for differentiating hashes. Multihash defines that by prefixing hashes with a hash function identifier and the hash size.
+
 [[def: Data Integrity]]
 
 ~ [W3C Data
--- a/spec/specification.md
+++ b/spec/specification.md
@@ -613,7 +613,7 @@ first [[ref: DID log entry]] and is the hash of the DID's inception event.
 To generate the required [[ref: SCID]] for a `did:tdw` DID, the DID Controller
 **MUST** execute the following function:
 
- `base58(multihash(JCS(preliminary log entry with placeholders)))`
+ `base58btc(multihash(JCS(preliminary log entry with placeholders)))`
 
 Where:
 
@@ -636,8 +636,8 @@ Where:
 3. `multihash` is an implementation of the [[ref: multihash]].
    Acceptable multihash identifier are defined in [[spec:controller-document]].
    Its output is a hash prefixed with a hash function identifier and the hash size.
-4. `base58` is an implementation of the [[ref: base58]] function.
-   Its output is the Base58 encoded string of its input.
+4. `base58btc` is an implementation of the [[ref: base58btc]] function.
+   Its output is the base58 encoded string of its input.
 
 ##### Verify SCID
 
@@ -671,7 +671,7 @@ previous log entry.
 ##### Generate Entry Hash
 
 To generate the required hash for a `did:tdw` DID entry, the DID Controller
-**MUST** execute the process `base58(multihash(JCS(entry)))` given a
+**MUST** execute the process `base58btc(multihash(JCS(entry)))` given a
 preliminary log entry as the string `entry`, where:
 
 1. `JCS` is an implementation of the [[ref: JSON Canonicalization Scheme]]
@@ -680,8 +680,8 @@ preliminary log entry as the string `entry`, where:
 2. `multihash` is an implementation of the [[ref: multihash]].
    Acceptable multihash identifier are defined in [[spec:controller-document]].
    Its output is a hash prefixed with a hash function identifier and the hash size.
-3. `base58` is an implementation of the [[ref: base58]] function.
-   Its output is the Base58 encoded string of its input.
+3. `base58btc` is an implementation of the [[ref: base58btc]] function.
+   Its output is the base58 encoded string of its input.
 
 The following is an example of a preliminary log entry that is processed to
 produce an entry hash. As this is a first entry in a DID Log, the input
@@ -711,7 +711,7 @@ Resolver **MUST** execute the following process:
 3. Set the first item of the entry to the `versionId` (first item) of the
    previous log entry. If this is the first entry in the log, set the value to
    the `1-<scid>` where `<scid>` if the SCID of the DID.
-4. Calculate the hash string as `base58(multihash(JCS(entry)))`, where:
+4. Calculate the hash string as `base58btc(multihash(JCS(entry)))`, where:
    1. `entry` is the data from the previous step.
    2. `JCS` is an implementation of the [[ref: JSON Canonicalization Scheme]]
       ([[spec:rfc8785]]). Its output is a canonicalized representation of its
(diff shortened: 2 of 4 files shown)


swcurran:
@bj-ms -- thanks!  Please add DCO sign-off to the PR. See DCO - Developer Certificate of Origin - https://github.com/apps/dco.  Instructions for fixing the commit are on the "Details" beside the failed check.

spec/specification.md
```diff
@@ -620,7 +620,7 @@ first [[ref: DID log entry]] and is a portion of the hash of the DID's inception
 To generate the required [[ref: SCID]] for a `did:tdw` DID, the DID Controller
 **MUST** execute the following function:
 
- `left(base32_lower(hash(JCS(preliminary log entry with placeholders))), <length>)`
+ `base58(hash(JCS(preliminary log entry with placeholders)))`
```
swcurran, line 623: Given that we want the flexibility/future-proofing to support different hash algorithms with backwards compatibility, I think we should use [multihash](https://github.com/dwyl/cid/issues/18) (at least) and perhaps [multibase](https://github.com/multiformats/multibase) here, so that a verifier can detect the hash (and encoding) algorithm(s) used.  The spec. will dictate the allowed hashes to generate, but a verifier should be flexible in receiving hashes.

Do we need multibase as well as multihash?

If we use multihash, the length of the (untruncated) SCID (and entry hashes) will be 46 characters -- [reference](https://github.com/dwyl/cid/issues/18).  So the SCID is up 16 from the previous 28, and we eliminate messing with the truncation.  I think that is fine -- especially once we get into PQ public keys :-)

spec/specification.md
```diff
@@ -643,10 +643,8 @@ Where:
 3. `hash` is the hash algorithm enumerated in the `hash` item in the [[ref:
    parameters]], or if none is specified, the default hash algorithm defined in
    this specification. Its output is the hash of its input.
-4. `base32_lower` is an implementation of the [[ref: base32_lower]] function.
-   Its output is the lower case of the Base32 encoded string of its input.
-5. `left` extracts the `<length>` number of characters from the string input.
-   1. `<length>` **MUST** be at least 28 characters.
+4. `base58` is an implementation of the [[ref: base58]] function.
```
swcurran, line 646: If we use multihash, I think we wind up using base58btc?  Or is it still base58?

spec/specification.md
andrewwhitehead: There's a minor security benefit to adding the hash algorithm, as it then could not be swapped in a replacement log file. That would still require producing a hash collision though, which is very unlikely.

Another option would be to add a version byte to the hash, fixing the generation algorithm version that was used. We're not currently using multihash so it would be a new dependency.

spec/specification.md
andrewwhitehead: The (expired) draft that is referenced is the Bitcoin version of base58, aka base58-btc.

spec/specification.md
```diff
@@ -47,8 +47,8 @@ here the full ABNF of those elements from that RFC would inevitably be wrong.
 tdw-did = "did:tdw:" scid ":" domain-segment 1+( "." domain-segment ) *( ":" path-segment )
 domain-segment = ; A part of a domain name as defined in RFC3986, such as "example" and "com" in "example.com"
 path-segment= ; A part of a URL path as defined in RFC3986, such as "path", "to", "folder" in "path/to/folder"
-scid = 28+( lower-base32 )
-lower-base32 = [2-7a-z]
+scid = ( base58 )
+base58 = [1-9A-Za-z]
```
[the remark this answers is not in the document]
swcurran: Or leave it off and leave it to the spec.

spec/specification.md
swcurran: Is there another spec we should be looking at?

spec/specification.md
brianorwhatever: I think we should use multibase and multihash and remove the `hash` parameter. Write normative statements saying which hash and base algs are allowed.

spec/specification.md
brianorwhatever: We should reference the VC controller document.

https://www.w3.org/TR/controller-document/#multibase-0
https://www.w3.org/TR/controller-document/#multihash

spec/specification.md
swcurran: I’m good with that.  We definitely want to constrain what a DID Controller is allowed to use (hence the normative statements) and allow a resolver to discover the DID Controller's choices.  I do think we keep the spec version and use that as an indicator to the resolver.  The spec. version dictates the options.

swcurran: So SCID calculation becomes:

`SCID = multibase(<base>,multihash(<hash>, JCS(input with placeholders)))`

brianorwhatever: yep version is fine I just want to remove `hash` from the parameters as that does the same job as multihash is.. other than the minor security benefit andrew mentions above which I don't see as a concern

andrewwhitehead: I can see multihash, but I'd rather not use multibase. We don't have a need to support multiple encodings and it's more work for verifiers.

spec/specification.md
swcurran: Side question @brianorwhatever — What is “Controller Document”?  There is no description of why it is and what it is for.  AFAICS, it is much like DIDs, and I heard Manu said it started with the DID Spec.

swcurran:
@bj-ms — are you good with updating your PR to reflect the feedback?  I think yours is close, but there will be a few more places to address.  Happy to do that as a follow up PR. Let me know how you want to proceed.

spec/specification.md
brianorwhatever: It's essentially where the normative statements for multibase/multihash had to be moved to in the VCWG as the group couldn't get consensus to have them in VCDM and there is a shared requirement between DI/JOSE.. I think basically a DID Document is a controller document

spec/specification.md
brianorwhatever: ok I'm fine with that. Also fine with not making the multihash change and just changing to base58btc in this PR

swcurran: I’d like to see multi-hash.  I can see a time where we add, for example, sha3-256 (as we briefly did), while still allowing sha-256 (since it is needed for long lasting DIDs). Multihash would be important.

Multibase less so. There are not going to be massive breakthroughs in efficiency or security related to encoding schemes :-).

bj-ms:
> @bj-ms — are you good with updating your PR to reflect the feedback? I think yours is close, but there will be a few more places to address. Happy to do that as a follow up PR. Let me know how you want to proceed.

I'm good with that, I will make the following modifications:

- Use multihash
- Use base58 mentioned in the VC spec
- Remove hash property from the did log
- Remove the definition of SCID in the chapter Method-Specific Identifier

spec/definitions.md
```diff
@@ -78,12 +81,12 @@ when forced to change (such as when an organization is acquired by another,
...
-[[def: base32_lower]]
+[[def: base58btc]]
 
-~ Applies [[spec:rfc4648]] to convert
-data to a `base32` encoding, and then lower cases the result. Data encoded as
-base32 consists of a string of characters containing only the letters A-Z and
-digits 2-7.
+~ Applies [[spec:draft-msporny-base58-03]] to convert
+data to a `base58` encoding. Data encoded as
+base58 consists of a string of characters containing only the letters A-Z, a-z and
```
brianorwhatever, line 88: base58 consists of `123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz` This purposely leaves off some characters that might be misread by humans.

spec/specification.md
```diff
@@ -680,17 +671,17 @@ previous log entry.
...
 1. `JCS` is an implementation of the [[ref: JSON Canonicalization Scheme]]
    ([[spec:rfc8785]]). Its output is a canonicalized representation of its
    input.
-2. `hash` is the hash algorithm enumerated in the `hash` item in the [[ref:
-   parameters]], or if none is specified, the default hash algorithm defined in
-   this specification. Its output is the hash of its input.
-3. `base32_lower` is an implementation of the [[ref: base32_lower]] function.
-   Its output is the lower case of the Base32 encoded string of its input.
+2. `multihash` is an implementation of the [[ref: multihash]].
+   Acceptable multihash identifier are defined in [[spec:controller-document]].
```
brianorwhatever, line 681: I think we should be explicit in saying we only allow sha256 in this version.

spec/definitions.md
swcurran: Suggest we don’t repeat the data here — just reference the spec.

spec/specification.md
swcurran: Here is should not reference specifically `sha256`, but should say something along the lines of “The hash algorithm selected **MUST** be as permitted by the version of the `did:tdw` specification being used for the log entry.”

bj-ms: We are currently missing a definition of the default hash function in the spec.  I think we should add it in the section where we define the generation of the SCID or in a disclaimer section about used crypto suites?

swcurran: It is defined where the “parameters” are defined.  It includes what parameters exist and where applicable, what enumerated values are allowed.  RIght now, that is where `hash` is defined.  Per recent discussions, on a soon to be made update, we’ll remove that parameter, and have the `model` parameter enumerate the permitted version values (currently `did:tdw:0.3`), and what each version implies WRT algorithms — hash, cryptosuite, etc.

martipos:
I am happy with the changes to base58 and skipping the trimming. Just two additional inputs. 

1. It seams base58 does not perform very well (in term of speed) with large byte values. Given that the SCID is generated based on the entry log this should however not be an issue as I assume the entry log will not reach a sized of concern (or are there any things planned on that end) ?
2. Out of curiosity, would double hashing be of interesting in generating the SCID?

swcurran:
The first issue is not a concern, as the input will always be a hash, which will be (for now) 32 bytes, perhaps longer if/when other hash algorithms are needed. Never more than 100 characters.

I’m not sure about a double hash as I don’t know how to calculate collision probability and how that translates into “time to calculate” in typical conditions (including Quantum conditions).  AFAIK, the attacker must generate a new log entry with a key they control and take over the web server publishing the file.  The hashing controls the probability of the first part of that.

swcurran (approved):

swcurran:
There are a couple of very minor clarifications I'd like to make to this (most of which are a bit tangential), but not worth holding up the merge. I think we have all agreed on the content.  Let's get this merged.

andrewwhitehead (approved):

andrewwhitehead:
The double hashing could be worth exploring if it increases the preimage resistance, although the difference would be pretty minor. Bitcoin uses it but without any real explanation of the purpose, IIRC. It would be easy to support without adding additional hash functions, but I'm not sure how it would compare to sha512/256 in strength.

PR #80, opened 2024-07-26
Align defs and refs

swcurran:
Mostly fixes the list of `defs` and corresponding `refs`.  Notes:

* For the most part, every text reference to a `def` **SHOULD** have `ref`, but there may be some missing. I used some scripting to do this (and then review), but there may be some missing.
* The `definitions.md` file has the most changes because I alphabetized the list. I didn’t (intend to) make any content changes.
* There should not be any refs in markdown Headers, `defs`, the definition text for a def (except to other `def`s), code blocks, etc.
* I tried to put no other changes in, but there are a couple that are clearly corrections — no content changes
* No change to the spec version, because there is no change to the text, just clarifications.
*  I did not add any new `defs`
* I removed `ref` tags on `DID`, `did:web` and `did:key`.  References to at least the two DID Methods will be added back in an orderly way in another PR — either a reference to their specs, or certain places beside where they are used in the text.

So — lots of changes, but not a lot to worry about. A sanity check would be nice, but we can fix anything that is missed.


spec/definitions.md
```diff
@@ -53,84 +13,106 @@ is a specification of mechanisms for ensuring the authenticity and integrity of
...
-DID subject can be found. The Decentralized Identity Foundation hosts the [Linked VP Specification](https://identity.foundation/linked-vp/).
+~ A `DID:tdw` entry hash is a hash generated using a formally defined process
+over the input data to a [[ref: log entry]], excluding the [[ref: Data Integrity]] proof
+item. The input data includes content from the predecessor to the version of the
+DID, ensuring that all the versions are "chained" together in a microledger. The
+generated [[ref: entry hash]] is subsequently included in the `versionId` element of the [[ref: log entry]] and **MUST** be
+verified by a resolver.
 
-[[def: Verifiable Credential, Verifiable Credentials]]
+[[def: ISO8601, [[ref: ISO8601]] String]]
```
brianorwhatever, line 70: This self-referential def intrigues me..

brianorwhatever (approved):
This looks good but it opens up a question to be solved in the future. What triggers a version update?

spec/definitions.md
swcurran: I have some “semi-automated tools”.  Messing with regex’s got me so far.  I thought I had that one (no replacements in lines starting with “[[def”), but alas…

Thanks!

brianorwhatever (approved):

PR #84, opened 2024-07-31
Clarifications of did:tdw features (abstract) and definitions

martipos:
references added, in the abstract and the definitions

SCID benefits summarized and description on portability added and separated from it, both in abstract

update of definitions for DID Log and Log Entries, and added definition for did:web

### DIFF
--- a/spec/abstract.md
+++ b/spec/abstract.md
@@ -4,22 +4,24 @@ Trust DID Web (`did:tdw`) is an enhancement to the `did:web` DID method,
 providing complementary web-based features that address `did:web`'s
 limitations. `did:tdw` features include:
 
-- Ongoing publishing of all DID Document (DIDDoc) versions for a DID instead of,
+- Ongoing publishing of all DID Document ([[ref: DIDDoc]]) versions for a DID instead of,
   or alongside a current `did:web` DID/DIDDoc.
 - The same DID-to-HTTPS transformation as `did:web`.
 - Supports the same [High Assurance DID-to-DNS] mechanism.
 - The ability to resolve the full history of the DID using a verifiable chain of
   updates to the [[ref: DIDDoc]] from genesis to deactivation.
-- A [[def: self-certifying identifier]] (SCID) for the DID that is globally
-  unique, embedded in the DID, and derived from the initial [[ref: DIDDoc]]. The SCID
-  enables [[ref: DID portability]], such as moving the DID's web location (and
-  so changing the DID string itself) while retaining a connection to the
-  predecessor DID(s) and the DID's verifiable history.
+- A [[def: self-certifying identifier]] (SCID) for the DID. The SCID, globally unique and
+  embedded in the DID, is derived from the initial [[ref: DIDDoc]]. It ensures the integrity
+  of the DID's history mitigating the risk of attackers creating a new object with
+  the same identifier.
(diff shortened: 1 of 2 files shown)


spec/abstract.md
```diff
@@ -4,22 +4,24 @@ Trust DID Web (`did:tdw`) is an enhancement to the `did:web` DID method,
...
 - The same DID-to-HTTPS transformation as `did:web`.
 - Supports the same [High Assurance DID-to-DNS] mechanism.
 - The ability to resolve the full history of the DID using a verifiable chain of
   updates to the [[ref: DIDDoc]] from genesis to deactivation.
-- A [[def: self-certifying identifier]] (SCID) for the DID that is globally
-  unique, embedded in the DID, and derived from the initial [[ref: DIDDoc]]. The SCID
-  enables [[ref: DID portability]], such as moving the DID's web location (and
-  so changing the DID string itself) while retaining a connection to the
-  predecessor DID(s) and the DID's verifiable history.
+- A [[def: self-certifying identifier]] (SCID) for the DID. The SCID, globally unique and
```
swcurran, line 13: This “def” should be a “ref” — my mistake.

spec/abstract.md
```diff
@@ -4,22 +4,24 @@ Trust DID Web (`did:tdw`) is an enhancement to the `did:web` DID method,
...
+  embedded in the DID, is derived from the initial [[ref: DIDDoc]]. It ensures the integrity
```
swcurran, line 14: Should be: “is derived from the initial [[ref: DID log entry]].”

spec/definitions.md
```diff
@@ -30,14 +30,15 @@ by the controller of the DID.
 
 [[def: DID Log, DID Logs]]
 
-~ A log of JSON arrays each of which provides the information necessary to
-generate and validate a version of the [[ref: DIDDoc]] from the previous version.
+~ A DID Log is a list of [[ref: Entries]] one being added for each update of an entry item,
+including new versions of the [[ref: DIDDoc]] or changed information necessary to generate or validate the DID.
 
-[[def: DID [[ref: Log Entry]], [[ref: DID Log]] Entries]]
+[[def: Entry, Entries, DID [[ref: Log Entry]], [[ref: DID Log]] Entries]]
```
swcurran, line 36: Arrgghh…those “refs” sneaking into the defs.  I’ll fix my “semi-automatic” tool to not do that in the future.  In the mean time, please remove the “ref” links in the “def” line.  Again — my mistake.

spec/definitions.md
```diff
@@ -56,7 +57,12 @@ providers.
 
 [[def: DID:web]]
```
swcurran, line 58: Should be lower case “DID” — my bad.

swcurran (changes requested):
Looks good — a few tweaks, mostly from things that I had wrong in the first place. What about also adding your name to the editors list?

spec/abstract.md
martipos: uu, this was a nasty one.

spec/definitions.md
martipos, line 58:
```suggestion
[[def: did:web]]
```

martipos:
messed up the DCO as the names used on github and locally seam not to be the same - thought they would be synced. my mistake. Will therefore close this PR and create a new one with the review comments addressed.

PR #85, opened 2024-08-01
Clarifications of did:tdw features (abstract) and definitions - V2

martipos:
re-do of RP#84 including addressed corresponding feedback. New RP was needed due to dco troubles - ignore the original RP#84

### DIFF
--- a/spec/definitions.md
+++ b/spec/definitions.md
@@ -30,14 +30,15 @@ by the controller of the DID.
 
 [[def: DID Log, DID Logs]]
 
-~ A log of JSON arrays each of which provides the information necessary to
-generate and validate a version of the [[ref: DIDDoc]] from the previous version.
+~ A DID Log is a list of [[ref: Entries]] one being added for each update of an entry item,
+including new versions of the [[ref: DIDDoc]] or changed information necessary to generate or validate the DID.
 
-[[def: DID [[ref: Log Entry]], [[ref: DID Log]] Entries]]
+[[def: DID Log Entry, DID Log Entries, Entries]]
 
-~ A DID Log Entry is a JSON array of items that define the authorized
+~ A DID Log Entry is a JSON array of five items which define the authorized
 transformation of a [[ref: DIDDoc]] from one version to the next. The initial entry
-establishes the DID and version 1 of the [[ref: DIDDoc]].
+establishes the DID and version 1 of the [[ref: DIDDoc]]. All entries are stored
+in the [[ref: DID Log]].
 
 [[def: DID Method, DID Methods]]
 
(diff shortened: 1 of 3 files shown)


spec/definitions.md
```diff
@@ -48,15 +49,21 @@ the DID Method Specification for `DID:tdw`.
...
 DID while retaining the [[ref: SCID]] and the history of the DID. This is useful
 when forced to change (such as when an organization is acquired by another,
 resulting in a change of domain names) and when changing DID hosting service
 providers.
 
-[[def: DID:web]]
+[[def: did:web]]
 
-~ `DID:web`...
+~ `DID:web` as described in the [W3C specification](https://w3c-ccg.github.io/did-method-web/)
```
brianorwhatever, line 60:
```suggestion
~ `did:web` as described in the [W3C specification](https://w3c-ccg.github.io/did-method-web/)
```

spec/definitions.md
```diff
@@ -48,15 +49,21 @@ the DID Method Specification for `DID:tdw`.
...
+is a DID method that leverages the Domain Name System (DNS) to perform the DID operations.
+It is valued for its simplicity and ease of deployment compared to DID methods that are
+based on distributed ledgers or blockchain technology, but also comes with increased
+challenges related to trust and security. DID:web provides a starting point for `DID:tdw`,
```
brianorwhatever, line 64:
```suggestion
challenges related to trust and security. `did:web` provides a starting point for `did:tdw`,
```

spec/definitions.md
```diff
@@ -48,15 +49,21 @@ the DID Method Specification for `DID:tdw`.
...
+which complements did:web with specific features to address the limitation of `DID:web`
```
brianorwhatever, line 65:
```suggestion
which complements `did:web` with specific features to address the limitations
```

brianorwhatever (approved):
Small casing suggestions but those could be addressed after the fact. This looks good enough to be merged to me

martipos:
as upper-/lowercase seems to be a thingy it may be helpful to define (at least for the few most prominent cases) when to use what. 

some suggestions and thoughts:
- did methods -> lowercase, e.g., "did:tdw", "did:web", "did:key"
- names -> upercase, e.g., "DID", "VC"
   => what about "DID Log", "DID Log Entry", "Entry" (in the sense of a DID Log Entry)  vs "DID log", "DID log entry", "entry" ?

brianorwhatever:
Those casings sound good to me

swcurran:
On the casing -- I went with DID Methods, DID and VC as you say.  For the DID Log Entry vs. DID log entry, I was a little more flexible, not requiring the capitalizing of them in most cases.  The ref/def references are case insensitive, so in most cases (I think) I left them as sentence-case.  However, I suspect I missed places.

Happy to define a fixed -- always capitalize them, or always use sentence case and I can do a pass over to fix the ones I messed up.

swcurran (approved):

PR #104, opened 2024-09-08
Replace parameters array with a parameters object in Generate a preliminary DID Log Entry section.

PatStLouis:
The spec currently instructs to generate the log entry as follows:
`When creating (registering) the DID the first entry starts with the follows items for processing: [ "{SCID}", "<current time>", "parameters": [ <parameters>], { "value": "<DIDDoc with Placeholders>" } ]`

However, in the examples the parameters section is an object and not an array. This PR looks at clarifying this step for implementers as I believe it's incorrect.

### DIFF
--- a/spec/specification.md
+++ b/spec/specification.md
@@ -225,8 +225,8 @@ Creating a `did:tdw` DID is done by carrying out the following steps.
 4. **Generate a preliminary DID Log Entry** (input JSON array)
    The [[ref: DID log entry]] is an input JSON array that when completed contains the following items:
    `[ versionId, versionTime, parameters, DIDDoc State, Data Integrity Proof ]`. When creating
-   (registering) the DID the first entry starts with the follows items for processing:
-   `[ "{SCID}", "<current time>", "parameters": [ <parameters>], { "value": "<DIDDoc with Placeholders>" } ]`
+   (registering) the DID the first entry starts with the following items for processing:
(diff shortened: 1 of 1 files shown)


andrewwhitehead:
I think it would be clearer if we just stated the values of the variables `versionId`, `versionTime` etc since the format of the line is laid out just before this.

PatStLouis:
@andrewwhitehead can you suggest an edit on my PR for this please?

PatStLouis:
I think the first value needs to remain as `{SCID}`

swcurran:
The data that you are talking about is the input to the SCID generation routine vs. the actual entry that will go into the DID log.  Is that clear?  I've struggled with getting that right.   The key differences between this and what goes into the log is the `versionId` is the literal `"{SCID}"`, and the parameters and `DIDDoc State` have the `"{SCID}"` placeholder where needed.  The `versionid` needs to be called out separately because it is not just a placeholder -- it differs considerably from the `<versionNumber>-<entryHash>` format.

brianorwhatever (approved):
This looks to fix a typo in the spec lets merge

swcurran:
@PatStLouis — can you update the base?  This will be replaced again with the pending array —> object change in the spec.

swcurran:
I think this issue is no overcome by events -- the new PR replaces the same line.

Closing -- please reopen or recreate once the 0.4 spec update is complete.

PR #118, opened 2024-10-16
Fix header link to the changelog

bj-ms:
The current link to the changelog in the header does not work because it does not match the changelog chapter id.

the link contains: https://identity.foundation/trustdidweb/#did-tdw-version-changelog

But it should contain: https://identity.foundation/trustdidweb/#didtdw-version-changelog


swcurran (approved):
Thanks — sorry about that.

PR #120, opened 2024-10-20
Add spec-up external spec

brianorwhatever:
I've set it up to use the same `external_specs` property that external refs use. It checks whether the spec maps to a string or object and does the external reference functionality for strings and this functionality for objects... probably not ideal and they can likely be the same thing I just didn't want to break the API for anyone currently using `external_specs`... You can take a look at the spec-up change that is required [here](https://github.com/brianorwhatever/spec-up/commit/fcab95edb4cc8d2a70f8c053e227220dbd6a08de)

### DIFF
--- a/spec/definitions.md
+++ b/spec/definitions.md
@@ -121,7 +121,7 @@ Specification](https://identity.foundation/linked-vp/).
 
 [[def: multihash]]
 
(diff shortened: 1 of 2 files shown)


swcurran:
Awesome — looks good.  Thanks!

Missed the DCO on the commit — can you fix that please, before we merge?

brianorwhatever:
@swcurran done

Issue #111, opened 2024-09-27
The ABNF for did:tdw seems to be missing the optional percent encoding of the port in the http url


daidoji:
> tdw-did = "did:tdw:" scid ":" domain-segment 1+( "." domain-segment ) *( ":" path-segment )
>domain-segment = ; A part of a domain name as defined in RFC3986, such as "example" and "com" in "example.com"
>path-segment= ; A part of a URL path as defined in RFC3986, such as "path", "to", "folder" in "path/to/folder"

Probably just needs a percent encoded optional ABNF snippet between the domain-segments and path-segments.

swcurran:
Good catch. Will add that after the current 0.4 PR is merged.

swcurran:
Action: Add this as a clarification without changing the version. @swcurran

Issue #110, opened 2024-09-27
It may be better to match the SubjectAltName (SAN) rather than the CommonName in the TLS certificate


daidoji:
I was reading through the new version of the spec and it jumped out at me:

> The domain name element of the method-specific identifier MUST match the common name used in the SSL/TLS certificate, and it MUST NOT include IP addresses.

Although the CN conceptually seems like the best field to check, I think Internet history and some ambiguity in how Netscape used the x509 specification way back in the day led to the modern browser behavior of checking the SAN as the authoritative domain identifier on a certificate.  Probably better to follow the browser convention for this did method unless there was a particular reason to choose CN validation for this did method that I don't understand.  (Although the most secure would be to do what's advised in the answer and check Subject, CN, and SAN.

  See this stack overflow and the RFCs mentioned in this answer for details: https://stackoverflow.com/a/5937270/488228

swcurran:
Interesting. Indeed, based on the description the SAN is the right way to check. I think what the requirement is really saying, although done so in a roundabout way, is that the URL **MUST** use the HTTPS protocol, and **MUST NOT** include IP addresses.   I would think it would be better to just say that, and not get into the details of how SSL/TLS certificates work.

Reasonable?

I’ll put in a PR after the current v0.4 PR is merged.  Or feel free to put one in if you want.

swcurran:
Ideally we point the SSL spec and remove the precision of the reference.  What we are trying to say is "Use SSL in the way modern browsers use SSL". @swcurran

PR #122, opened 2024-10-24
Clarifications about TLS Names and ABNF of ports

swcurran:
Addresses PRs #110 and #111.

Please review carefully the ABNF -- I think I have it right, but could use a second set of eyes on it.

@daidoji and @andrewwhitehead (and anyone else expert in ABNF).

### DIFF
--- a/spec/specification.md
+++ b/spec/specification.md
@@ -32,11 +32,12 @@ globally unique [[ref: SCID]] **MUST** be
 based on its initial content and placed into the DID identifier for publication
 and use.
 
-The domain name element of the method-specific identifier MUST match the
-common name used in the SSL/TLS certificate, and it MUST NOT include IP
-addresses. A port MAY be included and the colon MUST be percent encoded to
-prevent a conflict with paths. Directories and subdirectories MAY optionally be
-included, delimited by colons rather than slashes.
+The domain name element of the method-specific identifier MUST match the name
+found in the SSL/TLS certificate per [[spec:RFC6125]] and the its replacement
+[[spec:RFC9525]], and it MUST NOT include IP addresses. A port MAY be included
+and the colon MUST be percent encoded to prevent a conflict with paths.
+Directories and subdirectories MAY optionally be included, delimited by colons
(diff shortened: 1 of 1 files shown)


PatStLouis (dismissed):

spec/specification.md
```diff
@@ -45,8 +46,9 @@ examples below. The `domain-segment` and `path-segment` elements refer to
 here the full ABNF of those elements from that RFC would inevitably be wrong.
 
 ```abnf
-tdw-did = "did:tdw:" scid ":" domain-segment 1+( "." domain-segment ) *( ":" path-segment )
+tdw-did = "did:tdw:" scid ":" domain-segment 1+( "." domain-segment ) [ percent-encoded-port ] *( ":" path-segment )
 domain-segment = ; A part of a domain name as defined in RFC3986, such as "example" and "com" in "example.com"
+percent-encoded-port = "%3A" ( "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" ) 1*3( "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" )
```
daidoji, line 51: I think it should be `1*4` here as ports can go up to five digits.  Also there's a commonly used ABNF primitive called "DIGIT" you can use instead of listing all the values.

So its something like `["%3A"  1*5(DIGIT)]` should be fine if you'd rather do that.

swcurran: I was being picky :-).  The port (AFAIK) can’t have a leading 0 and “00000” can’t be used, hence the extra.

I didn’t know about 5 digit ports, but I guess that is obvious — 0 - 64k. So there you go.

Updating to: `percent-encoded-port = "%3A" ( "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" ) 1*4( DIGIT )`

What do you think?

swcurran: Updated.  Realized at the same time that the `scid` component was missing.

brianorwhatever (approved):

Issue #114, opened 2024-10-02
How to distinguish between witness proofs and controller proofs


brianorwhatever:
Currently the only difference between a witness proof and a controller's proof is that the witness will likely be a `did:tdw`. This isn't a particularly useful way to determine which is which. I think we need a better technique.. might be one reason for #66

swcurran:
As I was writing this — I think the answer is that the DID of the DID Controller **MUST** be a did:key, and **MUST** be in the authorized key array.  Conversely, witness did’s **MUST NOT** be `did:key`, and witness DIDs **MUST** be in the witness list.

A bit of data sniffing, but I don’t think outrageous.  The question is whether it is better or worse than using Proof Chains, given their lack of implementations?

swcurran:
@brianorwhatever -- do you want to see this added to the spec?  Seems it would be useful.

brianorwhatever:
For now what I am doing is checking if the verification method is a `did:tdw` or `did:key`. tdw is assumed to be a witness and key is assumed to be a controller

swcurran:
Action: As a clarification, without changing the version. @swcurran

swcurran:
Now separated into 2 files, so not an issue. Closing.

PR #123, opened 2024-10-24
Clarify that the DID Method of a witness DID and diffentiating the DID Log Entry proofs

swcurran:
Addresses #114 and clarifies that a witness DID **SHOULD** be a did:tdw and why.  Also mentions the possiblity of mutual witnesses causing an infinite loop.


spec/specification.md
```diff
@@ -895,6 +903,12 @@ An overview of the [[ref: witness]] mechanism is as follows:
   - Over time, the list of [[ref: witnesses]] may evolve, with each change being
     approved by the declared list of [[ref: witnesses]] from **before** such a
     change.
+  - The DID of the witnesses **SHOULD** be `did:tdw` DIDs to prevent the need for
```
brianorwhatever, line 906: shouldn't this be a MUST?

spec/specification.md
```diff
@@ -895,6 +903,12 @@ An overview of the [[ref: witness]] mechanism is as follows:
...
+    a `did:tdw` resolver having to resolve DIDs of other DID methods.
+  - `did:tdw` resolvers **SHOULD** account for the the possibility that the
```
brianorwhatever, line 908: how 🫠

swcurran: I would assume it is something like breaking recursion - record when you have seen a DID in the process, and break when you see the same DID again.  

That said, the spec. just has to point things out — it doesn’t have to do anything about it.  That’s why there are clever developers :-).

swcurran:
@brianorwhatever  -- updated the DID, changing both SHOULDs to MUSTs and cleaned up the infinite loop wording.  I think the second MUST is also needed as a resolver that gets into an infinite loop isn't very useful.

swcurran:
Given the planned redirection with witnesses per #135 -- closing this as overcome by events.

PR #124, opened 2024-10-25
Agenda after 20241024 Meeting

swcurran:


brianorwhatever (approved):
We should try to take proper attendance going forward

swcurran:
A good question for @kimdhamilton — recording the names of meeting attendees.  In most of the meetings I participate in, it is up to the attendees to add their name to the attendee list (and lately, it’s become that no one does…). In a couple of places, I’ve seen the host record the names.  What does DIF prefer/require?

kimdhamilton:
If the group is collab-editing the agenda.md file during the session (via hackmd), then I could see people adding their own names. Because that's rare, it usually defaults to: whomever is capturing the notes writes it down.

Note that we hope to improve this very soon through automation, i.e. just collect it from the zoom info.

PR #129, opened 2024-11-14
Modify the authorized key signature process when the pre-rotation fea…

bj-ms:
TLDR: Modify the authorized key signature process when the pre-rotation feature is enabled. Commited keys must be used directly in the next log entry.

Hi everyone,

During IIW 39b, I talked to Stephen Curran and Patrick St-Louis about a potential issue I see with key pre-rotation as it is stated now in the specification. There is a case that an attacker could create a malicious version of the DID Document because the commited `updateKey` is only enforced **AFTER** the publication of the log entry. Normally in the key pre-rotation concept, if you commit to `nextKeyHashes`, you must use it to sign the next log entry.

But the security risk is extremely low right now because you need to fulfill those two conditions:
-  The Attacker must have compromised the current `updateKey` (theft or PQC)
- The Attacker must known the public key behind the `nextKeyHashes` (theft or data leak)

Even if the risk is low, i would propose to fix the pre-rotation concept now. better safe than sorry :)

We will need to also adapt the examples and I would argue to jump to version 0.5, as it is a breaking change.


brianorwhatever:
The text you have added is how I was originally thinking about it. Maybe I had made assumptions about the specification that weren't true. The current implementation in `trustdidweb-ts` requires **ANY** changes to the DID to be authorized by the currently active `updateKeys` which is what it seems like this PR is adding for pre-rotation keys.. am I missing something?

bj-ms:
If you want a secure linked list of hashes, you will need to sign a new did log entry with the **previous** `updateKey`. If you don't "link" the entries with the previous entry, anyone could just create a new did log entry with a new `updateKey` and sign the entry with that same key.

The difference with pre-rotation keys, is that you commit to a new key with `NextKeyHash`. When you create a new did log entry, you need to sign with the new commited `updateKey` to use the full potential of pre-roration keys.

tldr: without pre-roration keys, you need to link with the signature of the **previous** `updateKey`. If you use pre-rotation keys, you link the did log entries with the `nextKeyHashes` and must use the **current** commited `updateKey` to generate the signature.

swcurran:
That is very confusing — I didn’t really understand the change based on the PR.   I think I now get the idea (questions, summary below), but I think it will take a bigger update to the specification to make it clear.

I think the core idea of what you are saying is that because of the pre-commitment, there is no need to wait until **after** “updateKeys” published to use them — we can sign the record with a key from the list in the **current** record.  

I don’t think it really reduces the threat significantly. From what you had above, we now have:

- The Attacker must known the public key behind the nextKeyHashes (theft or data leak)
- The Attacker must known the public key behind the nextKeyHashes key (theft or PQC)

That is only a slight variation on the risk that is motivating the change.

Is the following an accurate description of your proposal:

### No Prerotation:

- First entry is signed by a key in the first record “updateKeys” list.
- Subsequent entries are signed by a key in the **active** “updateKeys” list.
- An “updateKeys” list is **active** after the log entry it is in is written, through to when another “updateKeys” record is written (e.g., the new list becomes active).

### Preotation:

- There must by an “updateKeys” and “nextKeys” list in every record.
- All entries are signed by a key in the  “updateKeys” list of the **current** record.
- All of the keys in each “updateKeys” list **MUST** be in the **active** “nextKeys” list.
- A “nextKeys” list is active after the log entry it is in is written, through to when another “nextKeys” record is written (e.g., the new list becomes active).

**Question** — In your scheme, is it a requirement that the “updateKeys” and “nextKeys” list are in every record (first bullet above), or can multiple records be signed by keys in the same “updateKeys” list?

bj-ms:
Thanks for clarifying my description, yes it is accurate.

For your question, most other technical specification of pre-rotation I have seen forces the concept of `updateKeys` and `nextKeys` in each record. It is more secure because it closes the gap where attackers could compromise the **current** `updateKey` and sign a new log entry.

The main protection  that people want from pre-rotation keys is that even if the `updatekey` is compromised, the attacker cannot create a new log entry. And since the `nextKeyHash` is a hash, it is also more resistant to theorical PQC attacks.

I would recommend to force `updateKey` and `nextKey` in each record for security and simplicity.

bj-ms:
Decided to adapt the pre-rotation text with the global understanding discussed during the last meeting: https://hackmd.io/k4cIK9vQSlaeg2pdHE51IQ.

brianorwhatever:
I am still struggling to grok this so I have made diagrams to understand the process proposed AFTER this change has been accepted. Please correct me if I'm misunderstanding things. Of note: In the third image you see what I was concerned about  on the call which is the unused key **c** which is likely not an issue..
did_webh no presentation
did_webh prerotation 1
did_webh prerotation 2

bj-ms:
> I am still struggling to grok this so I have made diagrams to understand the process proposed AFTER this change has been accepted. Please correct me if I'm misunderstanding things. Of note: In the third image you see what I was concerned about on the call which is the unused key **c** which is likely not an issue.. did_webh no presentation did_webh prerotation 1 did_webh prerotation 2

The issue was that prerotation in version 1 was not working like you illustrated it. It should have worked like that but did not in the end.

To get the current prerotation version 1 you have to take the black arrows of version 0 and replace those in version 1.

My goal is to rewrite the prerotation parts to get to the actual version 1 you draw.
And Version 2 is the case when prerotation is enabled at a later stage and the green arrow is not necessary, as the transition starts after the first `nextKeyHashes` is set.

bj-ms:
@swcurran @brianorwhatever I added some clarifications in the specification. It should reflect pre-rotation version 1 and version 2 when pre-rotation is activated at a later stage from Brian's diagram.

swcurran (dismissed):
LGTM — nice work.  I think a nice addition would be the pictures that Brian put together in the comments. Worst case they go into the implementors section, but I think they could be useful in the spec. We can add them in a later PR for clarification.

swcurran:
@bj-ms -- I note that you didn't bump the version or add to the log.  Do you want to do that, or hold off?  I think that there should be at least a note in the version about the change.

We have at least two more things that need to go in that will bump the version -- the DID Method name change (#132 ) and updating the Witness proof handling (#135).

We could hold off on merging this until all three are ready and then merge them together such that the version number is preserved.  We also want to capture the v0.4 spec, as was done for the v0.3.

Thoughts?

bj-ms:
> @bj-ms -- I note that you didn't bump the version or add to the log. Do you want to do that, or hold off? I think that there should be at least a note in the version about the change.
> 
> We have at least two more things that need to go in that will bump the version -- the DID Method name change (#132 ) and updating the Witness proof handling (#135).
> 
> We could hold off on merging this until all three are ready and then merge them one immediately after the other such that we only change the version number once. We also want to capture the v0.4 spec, as was done for the v0.3.
> 
> Thoughts?

Good point, I will add to the log and bump to 0.5 and lets merge the name change and the new witness mechanism in a single version bump to 0.5

Once we have all changes in 0.5, we can prepare for 1.0

spec/specification.md
```diff
@@ -380,7 +380,7 @@ For each entry:
    DIDDoc]] for the version.
 8. If [[ref: Key Pre-Rotation]] is being used, the hash of all `updateKeys` entries
    in the `parameters` property **MUST** match a hash in
-   the active array of `nextKeyHashes` [[ref: parameter]], as defined in the
+   the  array of `nextKeyHashes` [[ref: parameter]] from the previous [[ref: DID log]] entry with exception of the first [[ref: DID log]], as defined in the
```
andrewwhitehead, line 383:
```suggestion
   the array of `nextKeyHashes` [[ref: parameter]] from the previous [[ref: DID log]]
   entry, with exception of the first entry, as defined in the
```

spec/specification.md
```diff
@@ -470,16 +470,21 @@ verifiable [[ref: DID Log Entry]] follows a similar process to the
 6. Generate a [[ref: Data Integrity]] proof on the [[ref: DID log entry]] using
    an authorized key, as defined in the [Authorized Keys](#authorized-keys)
    section of this specification.
-7. If the [[ref: DID Controller]] has opted to use [[ref: witnesses]] for the
+7. If [[ref: Key Pre-Rotation]] is being used, the hash of all `updateKeys` entries
+   in the `parameters` property **MUST** match a hash in
+   the  array of `nextKeyHashes` [[ref: parameter]] from the previous [[ref: DID log]] entry with exception of the first [[ref: DID log]], as defined in the
```
andrewwhitehead, line 475:
```suggestion
   the array of `nextKeyHashes` [[ref: parameter]] from the previous [[ref: DID log]] entry with exception of the first entry, as defined in the
```

swcurran:
@bj-ms -- can you please resolve the conflicts?  Sorry about that, but I think the "landing page" changes that were put in as #137 are helpful.  Thanks!

swcurran (approved):

Issue #135, opened 2024-11-22
New approach to recording/publishing Witnesses


swcurran:
@andrewwhitehead proposed (in this [HackMD Doc](https://hackmd.io/TlvkAGw6RMq0i9gPanU1jg)) a modified witness scheme.  We summarized and discussed the document and approach at [2024-11-21]() meeting and came to the following concensus:

1. Data model for the `witnesses` parameter remains the same (list of witnesses, threshold).
2. Instead of the witness proofs going into the log, they are stored in a separate file, `witness.json` -- an array of proofs from witness, including the DID Log entry version number (1, 2, ...) to which the proof applies. *Or should it be the `version_id`?*
3. The array contains only the last two proofs from each witness from which proofs are collected. When a witness proof is retrieved, if there are already two in the file from the witness, remove the oldest, and add the new one before writing the file.
4. The processing logic (excluding implementation details) is: Given an entry to be witnessed, verify that the `witness.json` file contains valid witness proofs from a threshold of current witnesses on the current or **newer** log entries.
5. Questions from the HackMD document and discussed on the call:
    1. Witness proofs can be published before the DID log with a new log entry is published. This eliminates the race condition of the DID Log being published without corresponding witness proofs. The "two proofs" per witness allows for the verification of at least one of the proofs of the witness when the `witness.json` is published before the corresponding log entry. The resolver would ignore the proof of an unpublished entry.
    2. What DID Method can a witness use?  A long discussion ensued, with follow up in Signal. The consensus was that only `did:key` will be supported. A note should be added to the specification that a mechanism for establishing the identity/reputation of the witness is outside the scope of the specification. If implementers need that functionality, they must define how to do that, such as including the witness DIDs in a Trust Registry.
        1. By limiting the DID type of a witness, we can drop the line in the specification about "preventing an infinite loop in resolving witnesses with mutual witnesses".
        2. The discussion went into a tangential area about the use of keys in other than the latest DID. Given the resolution of **not** using `did:webvh` for witnesses, it is not relevant to this discussion, but is relevant in other contexts. We'll continue that discussion.
     3. NOT DISCUSSED: Should the addition of witnesses only be permitted in the first entry?
         1. I propose that it can be added at any time after establishing the DID. It need not be in the first record -- a DID Controller may decide to add that to a DID after using the DID for sometime, and should not have to migrate to a new DID with a new SCID just to activate the feature.
     4. NOT DISCUSSED: Is there a use case for turning off witnessing?
[...]

PR #136, opened 2024-11-27
Update the specification of the witness process to put the witness proofs in a separate file

swcurran:
Adjusts the specification of witnesses to match the proposal from @andrewwhitehead, as discussed in #135 and at the [Work Item meeting of 20241121](https://github.com/decentralized-identity/trustdidweb/blob/main/agenda.md#meeting---21-nov-2024)

- Witness DIDs must be `did:key` DIDs.
- Witness proofs are collected by the DID Controller and published in the file `witness.json` that is beside the `did.jsonl` file
- The process for collecting/publishing witness proofs for a version update
- The process of verifying the witness proofs, and that there are sufficient proofs for each DID log entry

The information site has not be updated - a PR will follow.

This PR will be merged at the same time as the pre-rotation update and be included in a version 0.5 of the specification.


swcurran:
Question: In the case of the `did.jsonl` file going into the `.well-known` folder, I've defined that the `witness.json` file will also go there. This is different from the `<did>/path/to/file` DID URLs where the `.well-known/` is removed from the URL as part of the DID-to-HTTPS transformation.  Is that expected?

andrewwhitehead:
I do think it makes more sense for it to be alongside the `did.jsonl`, and not follow the normal path resolution rules, although I could be persuaded otherwise. Technically I think we're supposed to register `did.jsonl` as a `.well-known` path, and would then need to do the same for the witness file ([registry here](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml)). Also I did suggest `did-witness.json` so they'd be close to each other when sorted but either works.

spec/specification.md
```diff
@@ -60,7 +60,10 @@ this specification) element in `did:tdw` that is not in `did:web`. As specified
 in the [DID-to-HTTPS Transformation](#the-did-to-https-transformation) section
 of this specification, `did:tdw` and `did:web` DIDs that have the same fully
 qualified domain and path transform to the same HTTPS URL, with the exception of
-the final file -- `did.json` for `did:web` and `did.jsonl` for `did:tdw`.
+the final file -- `did.json` for `did:web` and `did.jsonl` for `did:tdw`. For
+`did:tdw` DIDs using [[ref: witnesses]], another file `witness.json` is also
+find (logically) beside the `did.jsonl` web server file. See the
```
andrewwhitehead, line 65:
```suggestion
found (logically) beside the `did.jsonl` web server file. See the
```

spec/specification.md
```diff
@@ -89,9 +92,15 @@ retrieve the [[ref: DID Log]].
...
-      `<did>/whois` or `<did>/path/to/file as defined in the section [DID URL
+   1. If the DID is using [[ref: witnesses]], an extra JSON file containing the
+      witness proofs for the [[ref: DID entries]] must be published and
+      retrieved during resolution. The URL for the extra file is defined by
+      replacing the `/did.jsonl` at the end of the [[ref: DID Log]] URL with
+      `/witness.json`.
+   2. When this algorithm is used for resolving a DID path (such as
+      `<did>/whois` or `<did>/path/to/file` as defined in the section [DID URL
       Handling](#did-url-resolution)), append the DID URL path instead.
+8. The content type for the `did.jsonl` file **MUST** be `text/jsonl`.
```
andrewwhitehead, line 103: This might be difficult to configure in some cases

swcurran: Do you have an alternate wording proposal (e.g. change to SHOULD?) or just drop it?

spec/specification.md
```diff
@@ -397,6 +408,11 @@ For each entry:
          `nextKeyHashes` list in the [[ref: parameters]].
       6. All other `did:tdw` processing configuration settings as defined by in the
          `parameters` object.
+10. If the `parameters` for any of the versions define that some or all of
+  the [[ref: DID Log entries]] must be witnessed, further verification of
+  the [[ref: witness]] proofs must be carried out, as defined in the [DID
+  Witnesses](#did-witnesses) section of this specification.
+11. If any verifications fail, discard the DID as invalid with an error message.
```
andrewwhitehead, line 415: It's probably acceptable for an individual witness verification to fail as long as the rules are still satisfied?

spec/specification.md
```diff
@@ -871,108 +890,143 @@ When processing other than the first [[ref: DID log entry]] where the
...
+- The [[ref: witnesses]] ***MUST** hold their own copy of the published [[ref:
+  DID Log]] prior to the version being witnessed.
+- Each [[ref: witness]] verifies the [[ref: DID Log Entry]], as defined by this
+  specification. If not, the [[ref: witnesses]] **MUST NOT** approve the [[ref: log entry]].
+- Each [[ref: witness]] determines (based on the governance of the ecosystem)
+  if they approve of the DID version update.
+  - The meaning of "approve" for any given implementation is outside the scope of this specification.
+- If the verification is successful and approval granted, the [[ref: witness]]
+  sends a [[ref: Data Integrity]] proof across the [[ref: DID log entry]] (minus
+  the `proof` item) to the [[ref: DID Controller]] signed by the [[ref: witness]]'s key.
```
andrewwhitehead, line 999: Maybe "following the procedure in §... for signing log entries"?

spec/specification.md
andrewwhitehead: Yes, I think SHOULD is probably strong enough

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: I set it to SHOULD on this and the one for `did.jsonl`, which I assume has the same issue.

spec/specification.md
swcurran: Good point...I'll adjust.

swcurran: The point of 11 (where you made the content) is a catch-all for all the verifications (`did:jsonl` file found, SCID calculation, version number is incremented, entryhash validates and so on and so on). I clarified the wording on that.  The "allow witness validations to fail if threshold met" is else where in the spec.

swcurran: But on that point about failed verifications...

Given that we are using DID:key for the verifications and the proof is across a static object, and was verified before it was inserted into the `did_witness.json` file, I don't think any of them can fail on resolution.  It's not like the DID can disappear like a did:webvh and the key not be found. 
The only reason the proofs in that file would "fail" for a resolver is already covered in the spec -- the witness proof file with new proofs is published before the DID Log with the new version is published.

andrewwhitehead: It does seem unlikely to fail, but I suppose there could be a copy/paste error or it could use an unsupported proof type. The file could even include multiple proof formats for different resolver versions at some point, like in an ECC/post-quantum transition.

spec/specification.md
swcurran: Added -- hopefully the wording is sufficient.

spec/specification.md
swcurran: OK -- add a tweak (at line 1023 of the specification.md file) that generalizes the point about ignoring proofs that reference a version not in the DID Log 

`Ignore any proofs that do not verify. For example, a `did-witness.json` file  **MAY** contain proofs of pending (unpublished) [[ref: DID Log entries]]. Such proofs **MUST** be ignored by resolvers.`

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
+  - The proof generation process of the witness **MUST** match the one used by
+   the DID Controller in generating its proof across the [[ref: DID log entry]],
```
brianorwhatever, line 1002:
```suggestion
   the DID Controller in generating its proof for the [[ref: DID log entry]],
```

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
+   as defined in the [Authorized Keys](#authorized-keys) section of this
+   specification. 
+- The DID Controller updates the [[ref: witnesses]]'s proofs in the
+  `did-witness.json` file with the received proofs, and publishes the updated file.
+  - The specification leaves to implementers how [[ref: witness]] proofs are
+    conveyed to the [[ref: DID Controller]].
+  - The [[ref: DID Controller]] **MAY** publish the updated `did-witness.json` file
+    as new witness proofs are added to the file.
+  - The [[ref: DID Controller]] **MUST** publish the update `did-witness.json` file
```
brianorwhatever, line 1011:
```suggestion
  - The [[ref: DID Controller]] **MUST** publish the updated `did-witness.json` file
```

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
+To do so, resolvers must:
+
+- Successfully complete the non-[[ref: witness]] verifications of the [[ref: DID Log]].
+- Verify the [[ref: witness]] proofs that reference [[ref: DID Log entries]] in the [[ref: DID Log]].
+  - Ignore any witness proofs that do not verify. For example, a `did-witness.json` file
+    **MAY** contain proofs of pending (unpublished) [[ref: DID Log entries]].
+    Such proofs **MUST** be ignored by resolvers.
+- For each [[ref: DID log entry]] requiring witnessing, the resolver **MUST**
+  verify that the `did-witness.json` file contains verified [[ref: witness]] proofs from a
+  threshold of active [[ref: witnesses]] across the current or **later** log entries. If
```
brianorwhatever, line 1028:
```suggestion
  threshold of active [[ref: witnesses]] of the current or **later** log entries. If
```

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
+active [[ref: witnesses]] have a threshold of active witnesses approving the [[ref: log entry]].
+To do so, resolvers must:
+
+- Successfully complete the non-[[ref: witness]] verifications of the [[ref: DID Log]].
+- Verify the [[ref: witness]] proofs that reference [[ref: DID Log entries]] in the [[ref: DID Log]].
+  - Ignore any witness proofs that do not verify. For example, a `did-witness.json` file
+    **MAY** contain proofs of pending (unpublished) [[ref: DID Log entries]].
+    Such proofs **MUST** be ignored by resolvers.
+- For each [[ref: DID log entry]] requiring witnessing, the resolver **MUST**
+  verify that the `did-witness.json` file contains verified [[ref: witness]] proofs from a
```
brianorwhatever, line 1027:
```suggestion
  verify that the `did-witness.json` file contains verified [[ref: witness]] [[ref: Data Integrity]] proofs from a
```

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
+`text/json`.
+
+The data model for the `did-witness.json` file is:
+
+```json
+{
+    [
+       {
+            "version_num": "1",
+            "witness": "did:key:z82LkvR3CBNkb...",
```
brianorwhatever, line 963: This property isn't technically required as the `proof` will contain the `verificationMethod` property

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
-in the [[ref: DID Log]], the [[ref: witnesses]] listed in that entry must
-approve the version, since there are no "prior" [[ref: witnesses]].
-
-The data model for the `witness` [[ref: parameter]] is as follows.
-The threshold design borrows from the [[ref: verifiable conditions]]
-specification.
-
-```json
-"witness" : {
-  "threshold": n,
```
brianorwhatever, line 944: Is this defined elsewhere now?

brianorwhatever (changes requested):
Is the `threshold` concept described somewhere? It looks like it was removed here but referred to a few times.

Only blocking change I'd like to see is the `version_num` -> `versionId` one.

It also stands out to me that if we didn't want to invent a data model for this the file could simply be an array of Data Integrity Proof objects. To link these to the DID log entry we would have an `id` on the base proof from the controller and then these witness proofs would use that as the `previousProof`. This is the proof chain approach we've talked about in the past. This would make the file look like this

```
[
  {
    "type": "DataIntegrityProof",
    "cryptosuite": "eddsa-jcs-2022",
    "created": "2020-11-05T21:28:14Z",
    "verificationMethod": "did:key:zk123412...#zk123412",
    "proofPurpose": "assertionMethod",
    "proofValue": "z6Qnzr5CG9876zNht8BpStWi8H2Mi7XCY3inbLrZrm955QLBrp19KiWXerb8ByPnAZ9wujVFN8PDsxxXeMoyvDqhZ",
    "previousProof": "urn:uuid:60102d04-b51e-11ed-acfe-2fcd717666a7"
  },
  {
    "type": "DataIntegrityProof",
    "cryptosuite": "eddsa-jcs-2022",
    "created": "2020-11-05T21:28:14Z",
    "verificationMethod": "did:key:zk543543...#zk543543",
    "proofPurpose": "assertionMethod",
    "proofValue": "zVbY8nQAVHMrXFkXJpmEcqdoDwLWxaqA3Q1geV64oey5q2M3XKaxup3tmzN4DRFTLVqpLMweBrSxMY2xHX5XTYVQe",
    "previousProof": "urn:uuid:60102d04-b51e-11ed-acfe-2fcd717666a7"
  }
]
```

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
+Where:
+
+- `version_num` is the prefix (prior to the literal `-`) of the `version_id` of the [[ref: DID log entry]] to which the proof applies.
+- `witness` is the DID of the witness.
+- `proof` is the proof of the [[ref: DID Log Entry]], excluding the `proof` item, identified by the `version_num`.
+
+Only the two most recent proofs from each [[ref: witness]] are retained in the
+`did-witness.json` file. When a new, verified proof from a [[ref: witness]] is received,
+the [[ref: DID Controller]] adds it to the file. If two other proofs from that
+[[ref: witness]] are found in the file, the oldest **MUST** be removed. The presence of a valid
```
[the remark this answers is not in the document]
swcurran: No -- I think **SHOULD** is good enough -- with the addition of the purpose is to keep the file size down.

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
+changed (`did.jsonl` to `did-witness.json`). The media type of the file **SHOULD** be
+`text/json`.
+
+The data model for the `did-witness.json` file is:
+
+```json
+{
+    [
+       {
+            "version_num": "1",
```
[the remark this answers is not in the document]
swcurran: Sounds good.  I think I would prefer to use this data structure vs. the DI Array.  I think it is a bit subtle how things link together.  I think we can merge this (after the change to version_id) and then debate the proof array idea.

spec/specification.md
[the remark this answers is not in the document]
brianorwhatever:
```suggestion
[[ref: witness]] are found in the file, the oldest **SHOULD** be removed. The presence of a valid
```

swcurran:
Added a definition of threshold and added refs for its use. The threshold definition and example is in the `Witnesses` section under `Witness Threshold Algorithm`. I also adjusted the example to be a little more complicated, to give more rationale about the weightings (witnesses with different weights).

spec/specification.md
swcurran: Fixed and other references to "across" when talking about a proof were changed as well.  I left "across" when talking about a hash, but perhaps that is also not right?

spec/specification.md
swcurran: Fixed

spec/specification.md
```diff
@@ -177,7 +177,7 @@ Each entry is a JSON object consisting of the following properties.
    specification.
 4. The JSON object `state` contains the [[ref: DIDDoc]] for this version of the
    DID.
-5. The JSON array `proof` contains a [[ref: Data Integrity]] proof calculated across
+5. The JSON array `proof` contains a [[ref: Data Integrity]] proof calculated for
```
brianorwhatever, line 180: sorry.. so pedantic about this phrasing

```suggestion
5. The JSON array `proof` contains a [[ref: Data Integrity]] proof created for
```

spec/specification.md
```diff
@@ -929,20 +929,49 @@ mechanism should be defined by the governance of the ecosystem, such as the
...
 
-The use of the threshold and weighted approvals (versus needing approvals from
+The `witness` element in a [[ref: parameters]] object of a [[ref: DID Log
+entry]] has the following data structure:
+
+```json
+
+"witness" : {
+  "threshold": n,
+  "selfWeight": n,
```
brianorwhatever, line 941: I think we've talked about this before... but do we really need `selfWeight`? They will always be signing their own entry.. It feels like added complexity for little gain as they could just decrease the threshold and not include their own proof in that calculation

spec/specification.md
```diff
@@ -959,7 +988,7 @@ The data model for the `did-witness.json` file is:
 {
     [
        {
-            "version_num": "1",
+            "version_id": "1-QmfGEUAcMpzo25kF2Rhn8L5FAXysfGnkzjwdKoNPi615XQ",
```
brianorwhatever, line 991: to match the rest of the spec

```suggestion
            "versionId": "1-QmfGEUAcMpzo25kF2Rhn8L5FAXysfGnkzjwdKoNPi615XQ",
```

spec/specification.md
```diff
@@ -970,16 +999,16 @@ The data model for the `did-witness.json` file is:
 
 Where:
 
-- `version_num` is the prefix (prior to the literal `-`) of the `version_id` of the [[ref: DID log entry]] to which the proof applies.
+- `version_id` is the `version_id` of the [[ref: DID log entry]] to which the proof applies.
```
brianorwhatever, line 1002:
```suggestion
- `versionId` is the `versionId` of the [[ref: DID log entry]] to which the proof applies.
```

spec/specification.md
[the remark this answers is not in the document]
swcurran: Fixed

spec/specification.md
```diff
@@ -871,108 +891,148 @@ When processing other than the first [[ref: DID log entry]] where the
...
+  - The specification leaves to implementers *how* the [[ref: log entry]] data is provided to the [[ref: witnesses]].
+- The [[ref: witnesses]] ***MUST** hold their own copy of the published [[ref:
+  DID Log]] prior to the version being witnessed.
+- Each [[ref: witness]] verifies the [[ref: DID Log Entry]], as defined by this
+  specification. If not, the [[ref: witnesses]] **MUST NOT** approve the [[ref: log entry]].
+- Each [[ref: witness]] determines (based on the governance of the ecosystem)
+  if they approve of the DID version update.
+  - The meaning of "approve" for any given implementation is outside the scope of this specification.
+- If the verification is successful and approval granted, the [[ref: witness]]
+  sends a [[ref: Data Integrity]] proof across the [[ref: DID log entry]] (minus
```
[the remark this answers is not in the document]
swcurran: Fixed -- the first change was  "creates and sends..."

spec/specification.md
swcurran: Doh... thanks!!!

spec/specification.md
swcurran: Fixed

swcurran:
Another tweak in the "Update DID" section. There was a conflict with @bj-ms 's recently merged PR -- now addressed. In doing that, I ordered the list of actions better to what needs to happen -- fully prepare the JSONL line, deal with witness, publish the updated DID Log file.

spec/specification.md
```diff
@@ -333,26 +345,26 @@ are each a JSON object with the following properties:
...
    [[ref: parameters]].
    - While all [[ref: parameters]] in the first [[ref: Log Entry]] take effect
-     immediately, some kinds of [[ref: parameters]] defined in later [[ref: entries]] only take
-     effect *after* that entry has been published. For example, rotating the
-     authorized keys to update a DID takes effect only *after* the entry in
-     which they are defined has been published.
+     immediately, some kinds of [[ref: parameters]] defined in later [[ref:
+     entries]] only take effect *after* that entry has been published. For
+     example, rotating the authorized keys (in the array `updateKeys`) to update
+     a DID takes effect only *after* the entry in which they are defined has
```
brianorwhatever, line 359: unless we're using prerotation? 😅

spec/specification.md
```diff
@@ -607,10 +626,11 @@ properties are defined below.
...
   [DID Witnesses](#did-witnesses) section of this specification.
-  - A `witness` property in the first [[ref: DID log entry]] is used to define the
-    [[ref: witnesses]] and necessary threshold for that initial [[ref: log entry]]. In all other
-    [[ref: DID log entries]], a `witness` property becomes active **after** the publication
-    of its entry -- meaning its [[ref: log entry]] **MUST** be witnessed by the most recent
+  - A `witness` property in the first [[ref: DID log entry]] is immediately
+    "active" and used to define the [[ref: witnesses]] and necessary [[ref: threshold]]
+    for witnessing the initial [[ref: log entry]]. In all other [[ref: DID log entries]],
+    a `witness` property becomes active **after** the publication of its entry
+    -- meaning its [[ref: log entry]] **MUST** be witnessed by active
```
brianorwhatever, line 633: I don't think we say anything about whether `witness` can be disabled after it's been enabled or is this forbidden?

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
-The data model for the `witness` [[ref: parameter]] is as follows.
-The threshold design borrows from the [[ref: verifiable conditions]]
-specification.
+The [[ref: witness]] process for a DID provides a way for collaborators to work
+with the [[ref: DID Controller]] to "witness" the publication of new versions of
+the DID. This specification defines the technical mechanism for using [[ref:
+witnesses]]. Governance and policy questions about when and how to use the
+technical mechanism are outside the scope of this specification.
+
+Witnesses can prevent a [[ref: DID Controller]] from updating/removing DID
```
brianorwhatever, line 910:
```suggestion
Witnesses can prevent a [[ref: DID Controller]] from updating/removing
```

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+versions of a DID without detection by the witnesses. [[ref: Witnesses]] are
+also a further mitigation against malicious actors compromising both a [[ref:
+DID Controller]]'s authorization key(s) to update the DID, and the [[ref: DID
+Controller]]'s web site where the [[ref: DID log]] is published. With both
+compromises, a malicious actor could take control over the DID by rewriting the
+[[ref: DID Log]] using the keys they have comprised. By adding [[ref:
```
brianorwhatever, line 916:
```suggestion
[[ref: DID Log]] using the keys they have compromised. By adding [[ref:
```

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+which the [[ref: DID Log]] is published.
+
+##### Witness Lists
+
+The list of DIDs that witness DID updates are defined in the `witness`
+parameter, as described in the [Parameters](#didtdw-did-method-parameters)
+section of this specification. Once the first `witness` parameter has been added
+to a version, there is always an active list of witnesses, and a [[ref: threshold]] of
+the active witnesses must provide verified proofs about an update before the
+update can be published. If the a DID version contains a new (replacement) list
```
brianorwhatever, line 929:
```suggestion
update can be published. If a DID version contains a new (replacement) list
```

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
-DID Log. For example, if there are three [[ref: witnesses]], each with a `weight` of 1,
-the [[ref: DID Controller]] with a `selfWeight` of 2, and a `threshold` of 4, the
-threshold will be met by two [[ref: witnesses]] approving the change, plus the [[ref: DID
-Controller]].
+- threshold: an integer that must be attained or surpassed by the sum of the witnesses weights for a DID log entry to be considered approved.
+- witnesses: an array of witnesses, each including the required fields:
+  - id: the DID of the witness. The DID **MUST** be a `did:key` DID.
+  - weight: an integer that is the weight given to this witness's approval
+
+##### Witness Threshold Algorithm
```
brianorwhatever, line 968:
```suggestion
- `threshold`: an integer that must be attained or surpassed by the sum of the witnesses weights for a DID log entry to be considered approved.
- `witnesses`: an array of witnesses, each including the required fields:
  - `id`: the DID of the witness. The DID **MUST** be a `did:key` DID.
  - `weight`: an integer that is the weight given to this witness's approval

##### Witness Threshold Algorithm
```

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+
+The use of the [[ref: threshold]] and weighted approvals (versus needing
+approvals from all [[ref: witnesses]]) is to prevent faulty [[ref: witnesses]]
+from blocking the publishing of a new version of the DID. To determine if the
+[[ref: threshold]] has been met, sum the `weight` integer of the received
+approvals and if it equal to or more than the `threshold`, the update can be
```
brianorwhatever, line 974:
```suggestion
approvals and if it is greater than or equal to the `threshold`, the update can be
```

spec/specification.md
brianorwhatever: I'm fine with it in there -- makes for easy comparisons to the parameter entry

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+       ...
+   ]
+}
+```
+
+Where:
+
+- `versionId` is the `versionId` of the [[ref: DID log entry]] to which the proof applies.
+- `witness` is the DID of the witness.
+- `proof` is the proof of the [[ref: DID Log Entry]] (excluding the `proof` item) identified by the `version_id`.
```
brianorwhatever, line 1011:
```suggestion
- `proof` is the proof of the [[ref: DID Log Entry]] (excluding the `proof` item) identified by the `versionId`.
```

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+[[ref: witness]] are found in the file, the oldest **SHOULD** be removed. The presence of a valid
+proof is an attestation from a [[ref: witness]] that the current **and all prior** versions
+of the DID have been verified and approved by that witness.
+
+##### Witnessing a DID Version Update
+
+The following process is used to witness a DID version update:
+
+- The [[ref: DID Controller]] prepares the full [[ref: DID Log Entry]] (including the
+  `proof` element) for the new version of DID, and shares it with the active [[ref: witnesses]].
```
brianorwhatever, line 1025:
```suggestion
  `proof` element) for the new version of the DID, and shares it with the active [[ref: witnesses]].
```

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+  - The specification leaves to implementers *how* the [[ref: log entry]] data is provided to the [[ref: witnesses]].
+- The [[ref: witnesses]] ***MUST** hold their own copy of the published [[ref:
+  DID Log]] prior to the version being witnessed.
+- Each [[ref: witness]] verifies the [[ref: DID Log Entry]], as defined by this
+  specification. If not, the [[ref: witnesses]] **MUST NOT** approve the [[ref: log entry]].
```
brianorwhatever, line 1030:
```suggestion
  specification. If not verified, the [[ref: witnesses]] **MUST NOT** approve the [[ref: log entry]].
```

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+- Each [[ref: witness]] determines (based on the governance of the ecosystem)
+  if they approve of the DID version update.
+  - The meaning of "approve" for any given implementation is outside the scope of this specification.
+- If the verification is successful and approval granted, the [[ref: witness]]
+  creates and sends a [[ref: Data Integrity]] proof for the [[ref: DID log entry]] (minus
```
PatStLouis, line 1035: Wouldn't just signing the versionId be sufficient since the versionId is a sort of hash of the entry? Meaning if the entry validation passes, a proof over the versionId only would be sufficient?

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+A `did:tdw` resolver **MUST** verify that all [[ref: DID Log entries]] that have
+active [[ref: witnesses]] have a [[ref: threshold]] of active witnesses approving the [[ref: log entry]].
+To do so, resolvers must:
+
+- Successfully complete the non-[[ref: witness]] verifications of the [[ref: DID Log]].
+- Verify the [[ref: witness]] proofs that reference [[ref: DID Log entries]] in the [[ref: DID Log]].
+  - Ignore any witness proofs that do not verify. For example, a `did-witness.json` file
+    **MAY** contain proofs of pending (unpublished) [[ref: DID Log entries]].
+    Such proofs **MUST** be ignored by resolvers.
+- For each [[ref: DID log entry]] requiring witnessing, the resolver **MUST**
```
brianorwhatever, line 1062: An entry in the past might have required witnessing at the time but those proofs have been discarded. This reads as though they need to verify those old proofs. Later in the paragraph it says "current or later" so I think it's saying the right things here it's just a bit misleading. Maybe we move the "current or later" text earlier in the paragraph?

brianorwhatever (approved):

spec/specification.md
```diff
@@ -881,107 +901,174 @@ When processing other than the first [[ref: DID log entry]] where
...
+```
+
+Where:
+
+- `versionId` is the `versionId` of the [[ref: DID log entry]] to which the proof applies.
+- `witness` is the DID of the witness.
+- `proof` is the proof of the [[ref: DID Log Entry]] (excluding the `proof` item) identified by the `version_id`.
+
+To limit the size of the file, only the two most recent proofs from each [[ref: witness]] are retained in the
+`did-witness.json` file. When a new, verified proof from a [[ref: witness]] is received,
```
PatStLouis, line 1014: Does it make sense to instead only keep the proofs about the two most recent log update entries?

spec/specification.md
brianorwhatever: hmm interesting.. I don't hate this idea. For implementation I was originally thinking I would just verify a proof set all at once so this would make it two separate verifications but I don't think that's a big deal.

PatStLouis: that's an interesting point, so you just match the elements based on the version Id and merge the proofs into 1 array

spec/specification.md
andrewwhitehead: The resolver may still need to check proofs from witnesses that aren't active anymore, and aren't signing the latest log entries.

spec/specification.md
PatStLouis: could we then just have an object of key/value where the key is the versionId and the value is a proof set?
```
{
     "version-1": [{ witness-1-proof },{ witness-2-proof },{ witness-3-proof }],
     "version-2": [{ witness-1-proof },{ witness-2-proof },{ witness-3-proof }],
     "version-3": [{ witness-1-proof },{ witness-2-proof },{ witness-3-proof }],
}
```

PatStLouis: my idea was having theses objects with a normal data integrity signature applied to them, not the whole entry
```
[
     {
          "versionId": "version-1",
          "proof": [{ witness-1-proof },{ witness-2-proof },{ witness-3-proof }]
      },
     {
          "versionId": "version-2",
          "proof": [{ witness-1-proof },{ witness-2-proof },{ witness-3-proof }]
      },
]

```

andrewwhitehead: Yes, that might be a good simplification. If we're using DI then it still needs a JSON object payload though, so possibly something like `{"versionId":...}`?

PatStLouis: witness proof for the same entry/update should definitely be grouped together as a set

spec/specification.md
swcurran: Cleaned up to avoid that as an example.  The same example is repeated in a later step in the process. I removed the later step entirely since it is redundant, moving the **MUST** to this step.  The first DRY change!!!

spec/specification.md
swcurran: To be added as an issue.

spec/specification.md
swcurran: Fixed

spec/specification.md
PatStLouis: Yea, because all the witness really need to do is say "yes, I verified/validated this entry and all is good". The `versionId` is derived from the entry itself and is part of the verification, so it would carry the same idea IMHO.

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: I tweaked the wording a little, but I think it is "clear enough" (although it is a bit tricky).  We can work on it further in later PRs.  I don't think it is wrong as it is.

spec/specification.md
swcurran: I'll add to the issue and we can debate it.

spec/specification.md
swcurran: I'll add this to an issue and we can debate it.

andrewwhitehead: We could be a bit more precise and say, for each witness:
- Keep the latest proof from among the previous versions
- Add the proof against the unpublished version, if there is one

This would mean if they didn't sign the latest transaction, there would only be one proof for that witness.

swcurran: Let’s also do this as an add on. I like it.  We might get a little more thorough and say that applies to all witnesses, so any removed witness never has more than one proof.  Gets a little nuanced. I’ll add an issue.

PatStLouis (approved):

brianorwhatever (approved):

PR #137, opened 2024-11-27
Capture v0.4 and add note about main version being in flux

swcurran:
Adds a snapshot of the v0.4 spec, and then mark the "main" spec as being "In flux", while we get the prerotation and witness adjustments completed, plus the name change.

- adds a comment at the top of the spec about the current state of the spec.
- snapshots the current spec to be the v0.4 version

Leaves the current version as "undefined", until we declare the version at the next stable point -- hopefully Real Soon Now.


swcurran:
Updated to make v0.4 the landing page for the specification, and the primary `spec` folder the "work in progress" version of the specification.  Added notes in the readme of how to do that for times when we are starting a breaking change update to the specification, and how to set it back when we are back to just having a single stable specification.

README.md
```diff
@@ -101,3 +101,31 @@ To create a snapshot of a version:
 - Repository: [https://github.com/decentralized-identity/trustdidweb](https://github.com/decentralized-identity/trustdidweb)
 
 ```
+
+## Handling Version Transitions
+
+In the lifecycle of the specification, there will be times when the version is stable, but clarifications may be added, and when new versions are being defined with breaking changes.  We use the Spec-Up feature to have multiple versions (described above) to enable that, but it gets a little tricky. Notably, we want the landing page for the specification to **always** be the current version of the specification, **and** we want all "in progress" work to be to made to the single, primary specification, so that GitHub holds the full evolution of the specification.  To enable that, we adjust as needed the `"outpath_path"` in the `specs.json` file to define what version of the spec is on the specification landing page -- the spec version where `"output_path"` is set to `"./"`).
```
andrewwhitehead, line 107:
```suggestion
In the lifecycle of the specification, there will be times when the version is stable, but clarifications may be added, and when new versions are being defined with breaking changes.  We use the Spec-Up feature to have multiple versions (described above) to enable that, but it gets a little tricky. Notably, we want the landing page for the specification to **always** be the current version of the specification, **and** we want all "in progress" work to be to made to the single, primary specification, so that GitHub holds the full evolution of the specification.  To enable that, we adjust as needed the `"output_path"` in the `specs.json` file to define what version of the spec is on the specification landing page -- the spec version where `"output_path"` is set to `"./"`).
```

andrewwhitehead (dismissed):

andrewwhitehead (approved):

PR #139, opened 2024-11-28
Add CORS details

brianorwhatever:
Text pulled from [here](https://w3c-ccg.github.io/did-method-web/#cross-origin-resource-sharing-cors-policy-considerations)

### DIFF
--- a/spec/security_and_privacy.md
+++ b/spec/security_and_privacy.md
@@ -43,6 +43,6 @@ See also:
 
 ### Cross-Origin Resource Sharing (CORS) Policy Considerations
 
-To support scenarios where DID resolution is performed by client applications running in a web browser, the file served for the [[ref: DID Log]] should be accessible by any origin. To enable this, the [[ref: DID Log]] HTTP response can be set to include the following header:
+To support scenarios where DID resolution is performed by client applications running in a web browser, the file served for the [[ref: DID Log]] needs to be accessible by any origin. To enable this, the [[ref: DID Log]] HTTP response MUST be set to include the following header:
 
 `Access-Control-Allow-Origin: *`
--- a/spec/security_and_privacy.md
+++ b/spec/security_and_privacy.md
@@ -40,3 +40,9 @@ See also:
 
 - [UTS-46](https://unicode.org/reports/tr46/)
 - [[spec:RFC5895]]
+
+### Cross-Origin Resource Sharing (CORS) Policy Considerations
+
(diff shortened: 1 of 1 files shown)


spec/security_and_privacy.md
```diff
@@ -40,3 +40,9 @@ See also:
 
 - [UTS-46](https://unicode.org/reports/tr46/)
 - [[spec:RFC5895]]
+
+### Cross-Origin Resource Sharing (CORS) Policy Considerations
+
+To support scenarios where DID resolution is performed by client applications running in a web browser, the file served for the [[ref: DID Log]] should be accessible by any origin. To enable this, the [[ref: DID Log]] HTTP response can be set to include the following header:
+
```
swcurran, line 47: “can” — should that be a “SHOULD” in spec language, or in this part of the spec is that not appropriate?

brianorwhatever: good call - maybe even a MUST?

swcurran:
I think it is a **MUST**.  I think that is a common use case…

spec/security_and_privacy.md
```diff
@@ -40,3 +40,9 @@ See also:
 
 - [UTS-46](https://unicode.org/reports/tr46/)
 - [[spec:RFC5895]]
+
+### Cross-Origin Resource Sharing (CORS) Policy Considerations
+
+To support scenarios where DID resolution is performed by client applications running in a web browser, the file served for the [[ref: DID Log]] should be accessible by any origin. To enable this, the [[ref: DID Log]] HTTP response can be set to include the following header:
```
brianorwhatever, line 46:
```suggestion
To support scenarios where DID resolution is performed by client applications running in a web browser, the file served for the [[ref: DID Log]] needs to be accessible by any origin. To enable this, the [[ref: DID Log]] HTTP response MUST be set to include the following header:
```

brianorwhatever:
added the MUST. this is ready now I think

spec/security_and_privacy.md
```diff
@@ -40,3 +40,9 @@ See also:
 
 - [UTS-46](https://unicode.org/reports/tr46/)
 - [[spec:RFC5895]]
+
+### Cross-Origin Resource Sharing (CORS) Policy Considerations
+
+To support scenarios where DID resolution is performed by client applications running in a web browser, the file served for the [[ref: DID Log]] needs to be accessible by any origin. To enable this, the [[ref: DID Log]] HTTP response MUST be set to include the following header:
```
andrewwhitehead, line 46:
```suggestion
To support scenarios where DID resolution is performed by client applications running in a web browser, the file served for the [[ref: DID Log]] needs to be accessible by any origin. To enable this, the [[ref: DID Log]] HTTP response MUST include the following header:
```

A bit less wordy

swcurran (approved):

PR #150, opened 2024-12-05
Fix up definitions and references

brianorwhatever:
There were a number of dangling definitions and unresolved references I have cleaned up

### DIFF
--- a/spec/definitions.md
+++ b/spec/definitions.md
@@ -20,7 +20,7 @@ verifiable, decentralized digital identities. A DID refers to any subject (e.g.,
 a person, organization, thing, data model, abstract entity, etc.) as determined
 by the controller of the DID.
 
-[[def: DID Controller, DID Controllers]]
+[[def: DID Controller, DID Controller's, DID Controllers]]
 
 ~ The entity that controls (create, updates, deletes) a given DID, as defined
 in the [[spec:DID-CORE]].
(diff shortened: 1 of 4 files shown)


spec/definitions.md
```diff
@@ -129,14 +135,6 @@ verify it. Although [[ref: multihash]] supports many hash algorithms, for
 interoperability, [[ref: DID Controllers]] **MUST** only use the hash algorithms defined
 in this specification as permitted.
 
-[[def: multi-sig, multisig]]
```
brianorwhatever, line 132: I removed this once because it wasn't being used but the right thing to do might be to reference it somewhere in the witness stuff? The thing is, the witness proofs aren't realllllly pure "multisig" so I'm not sure..

swcurran: Agree that multi-sig should not be there.  We are not using it in the spec, and should not be refering to it.

swcurran (approved):
LGTM. I wish the VS Code wrap would keep links together. Tricky since refs and defs are SpecUp thingies...

brianorwhatever:
Sadly the dangling ref/def log messages still stick around for the old versions of the spec.. I like the static spec version solution you've come up with but I do have my gripes. Merging these in..

swcurran:
Agreed.  Arguable that we could “clarify” those specs by at least cleaning up the ref/defs.  The other thing being more vigilant about getting those fixed when we “declare” a version.  I should have done it long ago — just didn’t get around to it.

Issue #152, opened 2024-12-06
Update the data structure and what is signed by witness proofs


swcurran:
The data model for the `did-witness.json` file is not correct in the spec. right now. Will correct it.

In addition it was agreed that we would:

- make the `versionId` the object that gets signed -- not the entry.
- The embedded `proof` is an array of proofs from witnesses signing the `versionId`
  - This matches the Data Integrity model of the proof being added to a data structure it is signing.
  - This also allows the entire set of witness proofs to be verified using only the data in the `did-witness.json` file.
  - This would mean that there is no way to know if the witness actually verified a new DID Entry before approving it, or did they just blindly signed the `versionId` value. **Is that OK?**
- When deleting proofs, if a `versionId` has no proofs associated with it, it can be deleted from the file.

Please confirm that this is right @andrewwhitehead @PatStLouis @brianorwhatever and others.  @brianorwhatever (or someone) -- please let me know the data model for the `did-witness.json` file.

brianorwhatever:
Yeah, that looks correct to me. I think this is the data model we agreed on

```
[
  {
    "versionId": "1-Qmba111111...",
    "proof": [{ ... }, { ... }]
  },
  {
    "versionId": "2-Qzmb222222...",
    "proof": [{ ... }, { ... }]
  }
]
```

Issue #151, opened 2024-12-06
Should we support turning off pre-rotation and if so, how


swcurran:
Discussed at the Work Item [Meeting of 2024-12-05](https://hackmd.io/k4cIK9vQSlaeg2pdHE51IQ?view#Meeting---05-Dec-2024).

Do we want to allow turning off pre-rotation once started?

We probably do want to allow prerotation to be turned off.  If so:

- Mechanism is to not include the `nextKeyHashes` array in the parameters.
- In that log entry, the `updateKeys` array **MUST** contain keys committed to in the prior `nextKeyHashes` (e.g., pre-rotation active rules apply).
- In the next log entry, the "no pre-rotation rules" apply, and the auth proof must use a key from the `updateKeys` from the previous entry (e.g. same keys as signed the previous entry).

swcurran:
@bj-ms -- what do you think?

bj-ms:
Makes it a bit more complex to handle those state transition but it gives more flexibility at time where they would just create a new did because they don't want to use pre-rotation anymore.

I would agree to that proposition, we just have to be cautious with the implementations and the test-suites to get it right.

brianorwhatever:
it makes me feel uneasy but I haven't come up with any objections yet. We are assuming if a DID's next keys get compromised the malicious actor can already takeover control right? It feels odd to be degrading security levels I think is where my uneasiness comes from.

swcurran:
I think if we are defending against a malicious actor that has the next keys (and web server access), they would probably keep the pre-rotation as it prevents the party “fixing” the log — they wouldn’t have the new “next keys”.

@PatStLouis — this is a reminder that a web server implementation needs to be careful about what it is using for authorization to publish a new log.  We would like to make the access to the publication mechanism different from the authorization to update the log.  Outside of the spec, but something to think about.

brianorwhatever:
![did_webh prerotation (1)](https://github.com/user-attachments/assets/db050f4e-509c-42d7-ac01-9d650fbed3b7)
Is this what disabling prerotation would look like? Note the key `c` is used for 2 proofs in a row. Is that an issue? It means if you want to rotate keys as well as change your update keys it would take 2 log entries.

swcurran:
Yes — thanks — we can use that diagram as well :-)

swcurran:
I don’t think that is an issue.  There are two things that a DID Controller can do:

[...]

Issue #149, opened 2024-12-04
Specify requirements on witness key types


andrewwhitehead:
I think it might be sufficient to specify that the Data Integrity proofs in the witness file must also use a cryptosuite allowed by the specification. We do say to use the same method as the controller uses when signing a log entry. And then the key algorithm used by the `did:key` would naturally be restricted to keys supported by that set of cryptosuites (currently just Ed25519).

Issue #144, opened 2024-11-29
Adjust the did-witness.json file to minimize the number of proofs


swcurran:
From @andrewwhitehead in reviewing #136.

We could be a bit more precise and say, for each witness:

- Keep the latest proof from among the previous versions
- Add the proof against the unpublished version, if there is one

This would mean if they didn't sign the latest transaction, there would only be one proof for that witness.

PR #159, opened 2024-12-11
Updates the pre-rotation and witnesses sections with feedback from the first set of updates

swcurran:
Adjusts the recently changes pre-rotation and witness section, addressing a number of issues:

- Issue #151 adds that pre-rotation can be turned off, how (empty list) and the implications
- Issue #152 updates the `did-witness.json` data model to make the `versionId` what is signed (also issue #142)
- Issue #141 #143 #144 clarifying language on when and why to remove proofs and when there will be 1 or 2 proofs from a witness. Tricky language!
- Issue #149 the type of Data Integrity cryptosuite that the witnesss must use
- Issue #140 how to disable the use of witnesses for a DID

The goal is that this PR should complete the changes needed in v0.5 -- including the renaming.

Please review carefully!


spec/specification.md
```diff
@@ -428,6 +428,24 @@ Document the full list of error codes that can be generated in resolving a DID.
 
 ##### Reading did:webvh DID URLs
 
+A `didLwebvh` resolver **MUST** resolve the [[spec:DID-Core]] `versionId` and
```
brianorwhatever, line 431:
```suggestion
A `did:webvh` resolver **MUST** resolve the [[spec:DID-Core]] `versionId` and
```

spec/specification.md
```diff
@@ -964,10 +1001,9 @@ where:
 The use of the [[ref: threshold]] and weighted approvals (versus needing
 approvals from all [[ref: witnesses]]) is to prevent faulty [[ref: witnesses]]
 from blocking the publishing of a new version of the DID. To determine if the
-[[ref: threshold]] has been met, sum the `weight` integer of the received
-approvals and if it equal to or more than the `threshold`, the update can be
-published. The calculation **MUST**  be executed by resolvers processing a
-[[ref: DID Log]].
+[[ref: threshold]] has been met, all participants **MUST** sum the `weight`
+integer of the received approvals. and if it equal to or more than the
```
brianorwhatever, line 1005:
```suggestion
integer of the received approvals. and if it is equal to or more than the
```

spec/specification.md
```diff
@@ -976,40 +1012,60 @@ all of the first three witness (`1+1+1`) providing an approving `proof`.
...
-Proofs from witnesses are retained in a separate file (`did-witness.json`) from the
-[[ref: DID Log]]. The same [DID to HTTPS
+Proofs from [[ref: witnesses]] are placed into a separate file
+(`did-witness.json`) from the [[ref: DID Log]]. The same [DID to HTTPS
 Transformation](#the-did-to-https-transformation) used for the [[ref: DID Log]]
 is used to locate the `did-witness.json` resource, with only the last element
-changed (`did.jsonl` to `did-witness.json`). The media type of the file **SHOULD** be
-`text/json`.
+changed (`did.jsonl` to `did-witness.json`). The media type of the file
+**SHOULD** be `text/json`.
```
brianorwhatever, line 1020:
```suggestion
**SHOULD** be `application/json`.
```

spec/specification.md
```diff
@@ -1019,23 +1075,20 @@ The following process is used to witness a DID version update:
   `proof` element) for the new version of the DID, and shares it with the active [[ref: witnesses]].
   - The specification leaves to implementers *how* the [[ref: log entry]] data is provided to the [[ref: witnesses]].
 - The [[ref: witnesses]] ***MUST** hold their own copy of the published [[ref:
-  DID Log]] prior to the version being witnessed.
+  DID Log]] prior witnessing a [[ref: DID Log entry]].
```
brianorwhatever, line 1078:
```suggestion
  DID Log]] prior to witnessing a [[ref: DID Log entry]].
```

spec/specification.md
```diff
@@ -1019,23 +1075,20 @@ The following process is used to witness a DID version update:
...
-   as defined in the [Authorized Keys](#authorized-keys) section of this
-   specification. 
-- The DID Controller updates the [[ref: witnesses]]'s proofs in the
-  `did-witness.json` file with the received proofs, and publishes the updated file.
+  creates and sends to the [[ref: DID Controller]] a [[ref: Data Integrity]]
+  proof signed using the [[ref: witness]]'s `did:key` DID.
   - The specification leaves to implementers how [[ref: witness]] proofs are
     conveyed to the [[ref: DID Controller]].
+- The [[ref: DID Controller]] **MUST** create or add the proof to the record for
+  the applicable `versionID` for the unpublished [[ref: DID log entry]].
```
brianorwhatever, line 1090:
```suggestion
  the applicable `versionId` for the unpublished [[ref: DID log entry]].
```

spec/specification.md
swcurran: I have the JSON Lines type as `text/jsonl`.  That's correct?

brianorwhatever: It's "correct" in that its the leading candidate. JSON of course has more eyeballs so this is what wikipedia says

The official [MIME type](https://en.wikipedia.org/wiki/MIME_type) for JSON text is application/json,[[26]](https://en.wikipedia.org/wiki/JSON#cite_note-27) and most modern implementations have adopted this. Legacy MIME types include text/json, text/x-json, and text/javascript.[[27]](https://en.wikipedia.org/wiki/JSON#cite_note-28) The standard filename extension is .json.[[28]](https://en.wikipedia.org/wiki/JSON#cite_note-29)

brianorwhatever:
So with the new `null` option for nextKeyHashes what about `undefined` aka the parameter is left off. This should definitely be a failure case right?

andrewwhitehead:
> So with the new `null` option for nextKeyHashes what about `undefined` aka the parameter is left off. This should definitely be a failure case right?

If a parameter is left off it should carry on with its old value, although in this case the update would be invalid because we require it to be changed.

My perspective was that parameters that have never been set default to `null` and they can be reset to `null` (since `undefined` isn't supported in JSON). Although we could also define default (implicit) values for each parameter.

swcurran:
Need to get to a final answer on this.

1. We could add to the spec that all parameters are initialized to null.
2. We could add the default to null **and** remove the `empty list “[]”` option for the array items, leaving only `null` — used initially and when reset.

As noted in the spec, a missing value in the parameters means “unchanged from the previous value”. As Andrew notes,  some parameters add rules of when that a missing value is unacceptable — e.g. parameters that **MUST** be in the first entry, and the pre-rotation parameters that **MUST** be in every entry when active.

After writing this, my leaning is 2 above.

swcurran:
Added commit to limit array parameters (`witness`, `nextKeyHashes`, `updateKeys`) to be deactivated by `null`, removing the option of using of empty list `[]`.  Also went through the parameters and formalized their default values where not previously explicit, and how to use `null` to deactive them (where needed).

swcurran:
Add a last tweak -- made explicit that they array parameters **MUST NOT** be set to an empty list.

spec/specification.md
```diff
@@ -597,11 +621,15 @@ properties are defined below.
...
+    [deactivate](#deactivate-revoke) section of this specification for more
+    details.
+  - `updateKeys` **MUST NOT** be set to an empty list `[]`.
 - `portable`: A boolean (`true` / `false`) indicating if the DID is portable and
   thus can be renamed to change the Web location of the DID.
   - The value can **ONLY** be set to `true` in the first [[ref: log entry]], the initial version of the DID.
-  - If not explicitly set in the first [[ref: log entry]], it is set to `false`.
-  - Once the value has been set to `false`, it cannot be set back to `true`.
+  - If not explicitly set in the first [[ref: log entry]], it **MUST** be set to `false`.
+  - Once the value has been explicitly to `false` in a [[ref: DID log entry]], it **MUST NOT** be set back to `true`.
```
brianorwhatever, line 632:
```suggestion
  - Once the value has been explicitly set to `false` in a [[ref: DID log entry]], it **MUST NOT** be set back to `true`.
```

spec/specification.md
```diff
@@ -597,11 +621,15 @@ properties are defined below.
...
+  - `updateKeys` **SHOULD** be set to `null` when deactivating the DID. See the
+    [deactivate](#deactivate-revoke) section of this specification for more
+    details.
+  - `updateKeys` **MUST NOT** be set to an empty list `[]`.
 - `portable`: A boolean (`true` / `false`) indicating if the DID is portable and
   thus can be renamed to change the Web location of the DID.
   - The value can **ONLY** be set to `true` in the first [[ref: log entry]], the initial version of the DID.
-  - If not explicitly set in the first [[ref: log entry]], it is set to `false`.
-  - Once the value has been set to `false`, it cannot be set back to `true`.
+  - If not explicitly set in the first [[ref: log entry]], it **MUST** be set to `false`.
```
brianorwhatever, line 631: I think this kind of reads funny as it's not really a MUST it is just a fact.
```suggestion
  - If not explicitly set in the first [[ref: log entry]], its value is set to `false`.
```

spec/specification.md
```diff
@@ -610,12 +638,21 @@ properties are defined below.
   - The process for generating the hashes and additional details for using [[ref: pre-rotation]] are defined in the
     [Pre-Rotation Key Hash Generation and Verification](#pre-rotation-key-hash-generation-and-verification)
     section of this specification.
-  - Once the [[ref: parameter]] `nextKeyHashes` has been set, the [[ref: Key Pre-Rotation]] feature becomes active and the property **MUST** be present in any future [[ref: DID log entry]].
+  - If not explicitly set in the first [[ref: DID Log entry]], its value **MUST** be `null`.
```
brianorwhatever, line 641: Same as above

```suggestion
  - If not explicitly set in the first [[ref: DID Log entry]], its value is `null`.
```

spec/specification.md
```diff
@@ -610,12 +638,21 @@ properties are defined below.
...
   - All [[ref: multikey]] formatted public keys added in a new `updateKeys` list **MUST** have their
     hashes listed in the `nextKeyHashes` list from the previous [[ref: DID log entry]].
   - A [[ref: DID Controller]] **MAY** put extra strings in the `nextKeyHashes`
     array that are not subsequently used in an `updateKeys` entry.
   - Any unused hashes in the prior `nextKeyHashes` are ignored.
+  - The value of `nextKeyHashes` **MAY** be set to `null` to deactivate
+    pre-rotation. For additional details about turning off [[ref: pre-rotation]]
+    see the [pre-rotation](#pre-rotation-key-hash-generation-and-verification)
+    section of this specification.
+  - `nextKeyHashes` **MUST NOT** be set to an empty list `[]`.
```
brianorwhatever, line 655:
```suggestion
  - `nextKeyHashes` **MUST** have at least 1 entry and **MUST NOT** be set to an empty list `[]`.
```

spec/specification.md
```diff
@@ -627,14 +664,19 @@ properties are defined below.
     a `witness` property becomes active **after** the publication of its entry
     -- meaning its [[ref: log entry]] **MUST** be witnessed by active
     `witnesses` from a **prior** [[ref: DID log]] entry.
-- `deactivated`: A JSON boolean that **SHOULD** be set to `true` when the DID is to
-  be deactivated. See the [deactivate (revoke)](#deactivate-revoke) section of
-  this specification for more details.
+  - If the `witness` property is not set in the first [[ref: DID log entry]],
+    its value **MUST** be null.
```
brianorwhatever, line 668:
```suggestion
    its value is null.
```

spec/specification.md
```diff
@@ -627,14 +664,19 @@ properties are defined below.
...
+  - `witness` **MUST NOT** be set to an empty list `[]`.
```
brianorwhatever, line 669:
```suggestion
  - `witness` **MUST** have at least 1 entry and **MUST NOT** be set to an empty list `[]`.
```

spec/specification.md
```diff
@@ -627,14 +664,19 @@ properties are defined below.
...
+- `deactivated`: A JSON boolean that **MUST** be initialized to `false` and
```
brianorwhatever, line 670:
```suggestion
- `deactivated`: A JSON boolean that is initialized to `false` and
```

spec/specification.md
```diff
@@ -627,14 +664,19 @@ properties are defined below.
...
+  **SHOULD** be set to `true` when the DID is to be deactivated but remains
```
brianorwhatever, line 671:
```suggestion
  **MUST** be set to `true` when the DID is to be deactivated but remains
```

brianorwhatever (dismissed):
All of my comments are of the  "take it or leave it" type

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: I think leaving it as being testable makes sense.  Leaving this one.

spec/specification.md
swcurran: No change

spec/specification.md
swcurran: Fixed in all three places.

spec/specification.md
swcurran: No change

spec/specification.md
swcurran: Fixed

spec/specification.md
swcurran: No change

spec/specification.md
swcurran: Left as is. In the deactivate section we left it.  If a DID is "deactivated" by removing the JSONL file, for example, no need to set deactivate to true.  I wanted to leave that open.

brianorwhatever (approved):

PR #164, opened 2025-01-07
Fixes bad links in the spec header and updates the readme for spec version handling

swcurran:
Fixes some bad links in the spec landing page and updates to the readme to define how to handle version changes -- the cause of the bad links.

Also resolves a couple of ref/def errors noted in testing the changes.


brianorwhatever (approved):
LGTM

PR #168, opened 2025-01-17
Correct DID URL examples and description in v0.4, v0.5 and Editors Draft versions

swcurran:
This PR does two things that are corrections to mistakes in the v0.4/v0.5/Editor's Draft versions of the spec.

The first correction is that in the `DID URL Resolution` section, the example DIDs were not updated to reflect the change to the location of the SCID from v0.3. This was a mistake in the update of the specification, so this is a fix.

The second correction is that there is a contradiction between the `DID URL Resolution` introduction text and the `whois LinkedVP Section`. In the first, the text says the `whois.vp` file **is** to be placed in the `.well-known` folder if that is where the `did.jsonl` file is placed, while the second example says the file **is NOT** to be placed in the `.well-known` folder. The intention of the Editors was the latter, so the text in the DID URL section was corrected.

### DIFF
--- a/spec-v0.4/specification.md
+++ b/spec-v0.4/specification.md
@@ -1003,30 +1003,25 @@ The risk of publishing the `did:web` in parallel with the `did:tdw` is that the
 added security and convenience of using `did:tdw` are lost.
 
 ### DID URL Resolution
-
 The `did:tdw` DID Method embraces the power and usefulness of DID URLs, along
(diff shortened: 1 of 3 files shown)


brianorwhatever (approved):

swcurran:
@andrewwhitehead — does this match your implementation?  I wanted to be sure you saw this mistake in the spec.

Issue #51, opened 2024-04-29
Implementor Guidance: Publishing AnonCreds objects using did:webvh


swcurran:
This will not be part of the spec, or the implementors guide, but we will want a "did:tdw AnonCreds Method", and this issue is to get some ideas down on how to do that.

For most of the objects, the publishing mechanism will be:

- Use a similar method to the DID URLs in `did:indy` as defined in the [did:indy specification](https://hyperledger.github.io/indy-did-method/#did-urls-for-indy-object-identifiers). "Similar" except for the reference to the Indy Transaction ID of the Schema in dependent objects. This means the schemas are:
   - Schema: `<did>/anoncreds/v0/SCHEMA/<schema_id>/<schema_ver>.json`
   - CredDef: `<did>/anoncreds/v0/CREDDEF/<schema_id>/<schema_ver>/<cred_def_id>.json`
   - RevRegDef: `<did>/anoncreds/v0/REV_REG_DEF/<schema_id>/<schema_ver>/<cred_def_id>/<rev_reg_id>.json`

We propose that each object be a W3C VC signed by the DID, with the `credentialSubject` being the AnonCreds object.

That is all pretty straight forward. In theory, AnonCreds objects on Indy can be updated, although in practice that is not done as being unnecessarily complex -- no one has ever (AFAIK) implemented them. For the `did:tdw` AnonCreds Method, an update could overwrite the JSON file with new version or perhaps we could on publishing save two files, one with the `_<version>.json` (or `_<time>.json`), and one with just `.json`, so that the history was available.

Not listed above is the `RevRegEntry` object, which is a bit more interesting. In that case, it is updated each time the Issuer revokes a batch (1 or more) of credentials in the RevReg. That complexity must be handled. As defined in the [AnonCreds v1 Specification](https://hyperledger.github.io/anoncreds-spec/#anoncreds-revocation-data-flow) RevRegEntry contains:

- An ID
- A timestamp (UNIX Time - `epoch`)
- An accumulator for the used by the holder to create a Non-Revocation Proof (NRP) and by the verifier to verify the NRP
- Either the full state of the credentials (a bit per credential - revoked or not) or the delta of states since the last change.
   - For the `did:tdw` AnonCreds Method, we will use the "full state" method.
   - Might be worth looking at the Cheqd AnonCreds Method for how that was done, and to be consistent.
   - Presumably, a compressed bit string such as is used in StatusList2021.
[...]

PR #174, opened 2025-02-10
Add did:webvh AnonCreds Method specification

swcurran:
Adds a separate specification for the `did:webvh` AnonCreds Method. The spec defines `Attested Resources` and then specifies how they are used to wrap AnonCreds objects, define identifiers for the resources, and publish them.  Also covered is how each object type's identifier is found by those needing to resolve them -- holders and verifiers. This includes the list of RevRegEntries -- the state of a revocation registry each time an Issuer revokes a batch of credentials in a revocation registry.

### DIFF
--- a/spec-anoncreds-method/abstract.md
+++ b/spec-anoncreds-method/abstract.md
@@ -1,0 +1,10 @@
+## Abstract
+
+This `did:webvh` AnonCreds Method specification defines how [[spec: AnonCreds]] objects should be registered (written) and resolved when rooting them in a `did:webvh` DID. This specification parallels other DID-specific AnonCreds methods that are registered in the [AnonCreds Methods Registry].
+
+[AnonCreds Methods Registry]: https://hyperledger.github.io/anoncreds-methods-registry/
+
+The specification introduces the concept of an [[ref: Attested Resource]] that enables verifiable identifiers that resolve to resources (files) managed by a `did:webvh` Controller. We anticipate later extracting the [[ref: Attested Resources]] concept into a separate specification on its own. Each of the published AnonCreds objects ([[ref: schema]], [[ref: CredDef]], [[ref: RevRegDef]], and [[ref: RevRegEntry]] are treated as [[ref: Attested Resources]] in this AnonCreds method.
+
+For information beyond this `did:webvh` AnonCreds Method specification, the (`did:webvh`) DID method, and how (and
+where) it is used in practice, please visit [https://didwebvh.info/](https://didwebvh.info/).
--- a/spec-anoncreds-method/specification.md
+++ b/spec-anoncreds-method/specification.md
@@ -1,0 +1,376 @@
+## `did:webvh` AnonCreds Method Specification
+
+### Overview
+
+The `did:webvh` AnonCreds Method defines how AnonCreds objects ([[ref: Schemas]], [[ref: CredDefs]], [[ref: RevRegDefs]], and [[ref: RevRegEntries]]) are registered (published, written) by a [[ref: DID Controller]] using a `did:webvh` DID, and how others can resolve and verify those objects. The method makes use of [[ref: Attested Resources]], resources (objects, files) that are generated with a deterministic ID, attested to by the [[ref: DID Controller]], and published at a resolvable web location -- typically (but not necessarily) to the same web server as the [[ref: DID Controller]]'s `did:webvh` DID Log. The next section of the specification focuses on the publication and resolution of the AnonCreds objects.
+
+The specification of an [[ref: Attested Resource]] is found [later in this specification](#attested-resources) and may later be extracted into a standalone specification.  For those unfamiliar with [[ref: Attested Resources]], here is a brief summary of their important attributes:
+
+- [[ref: Attested Resources]] are a JSON structure that embed a complete resource (e.g., an AnonCred object, an encoded image, a [Bitstring Status List](https://www.w3.org/TR/vc-bitstring-status-list/), an [OCA Bundle](https://bcgov.github.io/aries-oca-bundles/), or any other resource published by the [[ref: DID Controller]]), plus metadata about the resource.
+  - Note that a `did:webvh` using [[ref: DID Controller]] does not have to publish resources as [[ref: Attested Resources]]. It is just a usefully verifiable data model.
+- The identifier from an Attested Resource is resolvable and contains a deterministic hash of the resource, allowing a resolver to verify that the content of the identified resource has not been altered.
+- The Attested Resource includes a [[ref: Data Integrity]] proof signed by the [[ref: DID Controller]] of the Attested Resource JSON (the resource, and metadata).
+- A `did:webvh` [[ref: Attested Resource]] is typically found at a relative path on the DID's web server. However, the DID's DIDDoc **MAY** use a DID `service` ([[spec:DID-CORE]]) to indicate the resources are based on a different URI.
+
+### AnonCreds Objects as Attested Resources
+
+Each `didwebvh` rooted AnonCreds object is the resource in an [[ref: Attested Resource]] published to the `did:webvh`. The identifier for each resource consists of three parts -- `<did>/path/to/resource/<hash>`, as follows:
+
+- `<did>`: The `did:webvh` DID.
+- `/path/to/file`: The location relative a base URI of a DIDDoc service where the [[ref: Attested Resource]] is located. Unless explicitly overridden in the `did:webvh` DIDDoc, the base URI is defined by the [`did:webvh` DID-to-HTTP Transformation].
+- `<hash>`: The deterministic hash of the [[ref: Attested Resource]] (based on this [generation algorithm](#creation-and-publishing)).
+
+[`did:webvh` DID-to-HTTP Transformation]: https://identity.foundation/didwebvh/#the-did-to-https-transformation
+
+Each `did:webvh` AnonCreds object is created by an AnonCreds verifiable credential Issuer, used to generate an [[ref: Attested Resource]], and published. Generating the [[ref: Attested Resource]] produces the object's identifier, and that identifier is in turn embedded into other objects (such as issued verifiable credentials) available to parties (Holders, Verifiers) that need to resolve the object.
+
+The following sections define how each AnonCreds object type is published as an [[ref: Attested Resource]].
+
+#### AnonCreds Schema
+
+An [[ref: AnonCreds Schema]] is the resource embedded in an [[ref: Attested Resource]]. The resulting [[ref: schema]] identifier is placed into an Issuer's [[ref: AnonCreds CredDef]] based on the [[ref: Schema]]. Holders and Verifiers resolve the [[ref: schema]] identifier retrieved from a resolved [[ref: AnonCreds CredDef]].  The `resourceType` **MUST** be `anonCredsSchema`. The `resourceName` **SHOULD** be the `name` of the [[ref: Schema]].
+
+The following is an example [[ref: AnonCreds Schema]] [[ref: Attested Resource]], with the identifier: `did:webvh:QmdhkLDvMUxSmf8LweLPergW5refqjL1S1YKpBihwvmXs6:id.anoncreds.vc:demo:8d2bebfe-6492-4915-b53c-4c3bf2c9711a/resources/zQmSPbRK7h8SCQKMXyNvtHZuSQnqL6yCCQ8UMe2Rfa4ucP9`
+
+```json
+{
+  "@context": [
+    "https://w3id.org/security/data-integrity/v2"
+  ],
+  "type": [
+    "AttestedResource"
+  ],
+  "id": "did:webvh:QmdhkLDvMUxSmf8LweLPergW5refqjL1S1YKpBihwvmXs6:id.anoncreds.vc:demo:8d2bebfe-6492-4915-b53c-4c3bf2c9711a/resources/zQmSPbRK7h8SCQKMXyNvtHZuSQnqL6yCCQ8UMe2Rfa4ucP9",
+  "content": {
+    "issuerId": "did:webvh:QmdhkLDvMUxSmf8LweLPergW5refqjL1S1YKpBihwvmXs6:id.anoncreds.vc:demo:8d2bebfe-6492-4915-b53c-4c3bf2c9711a",
+    "attrNames": [
+      "attributeClaim",
+      "predicateClaim"
+    ],
+    "name": "Demo Credential",
+    "version": "1.0"
+  },
+  "metadata": {
+    "resourceId": "zQmSPbRK7h8SCQKMXyNvtHZuSQnqL6yCCQ8UMe2Rfa4ucP9",
+    "resourceType": "anonCredsSchema",
+    "resourceName": "Demo Credential"
+  },
+  "proof": {
+    "type": "DataIntegrityProof",
+    "cryptosuite": "eddsa-jcs-2022",
+    "proofPurpose": "assertionMethod",
+    "proofValue": "z5ZTaBaz8pQz2Ne5S2aNAJW2K41eZmbXNqiyhNK3ejXiobAjVpYbKN3NLPVj9bXeYJdkA8Tfw7hbE5kTr52QxGzjK",
+    "verificationMethod": "did:webvh:QmdhkLDvMUxSmf8LweLPergW5refqjL1S1YKpBihwvmXs6:id.anoncreds.vc:demo:8d2bebfe-6492-4915-b53c-4c3bf2c9711a#key-01"
+  }
+}
+```
+
+#### AnonCreds CredDef
+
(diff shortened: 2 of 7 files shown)


swcurran:
@PatStLouis -- did I get it right?

swcurran:
Fixes #51

spec-anoncreds-method/specification.md
```diff
@@ -0,0 +1,376 @@
...
+- The identifier from an Attested Resource is resolvable and contains a deterministic hash of the resource, allowing a resolver to verify that the content of the identified resource has not been altered.
+- The Attested Resource includes a [[ref: Data Integrity]] proof signed by the [[ref: DID Controller]] of the Attested Resource JSON (the resource, and metadata).
+- A `did:webvh` [[ref: Attested Resource]] is typically found at a relative path on the DID's web server. However, the DID's DIDDoc **MAY** use a DID `service` ([[spec:DID-CORE]]) to indicate the resources are based on a different URI.
+
+### AnonCreds Objects as Attested Resources
+
+Each `didwebvh` rooted AnonCreds object is the resource in an [[ref: Attested Resource]] published to the `did:webvh`. The identifier for each resource consists of three parts -- `<did>/path/to/resource/<hash>`, as follows:
+
+- `<did>`: The `did:webvh` DID.
+- `/path/to/file`: The location relative a base URI of a DIDDoc service where the [[ref: Attested Resource]] is located. Unless explicitly overridden in the `did:webvh` DIDDoc, the base URI is defined by the [`did:webvh` DID-to-HTTP Transformation].
```
brianorwhatever, line 20: is the [`did:webvh` DID-to-HTTP Transformation] meant to be a link to the webvh spec section?

spec-anoncreds-method/specification.md
```diff
@@ -0,0 +1,376 @@
...
+
+Unlike the other objects, the [[ref: RevRegDef]] [[ref: Attested Resource]] also contains required metadata in the [[ref: Attested Resource]] `links` field that is outside of the [[ref: RevRegDef]] itself. Notably, `links` contains the list of all [[ref: RevRegEntry]] timestamps and identifiers associated with the [[ref: RevRegDef]] (see the example below). Each time, a new [[ref: RevRegEntry]] is published by the Issuer, a new [[ref: RevRegDef]] must be created that adds the new [[ref: RevRegEntry]] to the `links` list. Since the `links` item is not part of the resource (the [[ref: RevRegDef]]) of the [[ref: AttestedResource]], the [[ref: RevRegDef]] identifier (and more specifically, the hash within the identifier) does not change. The [[ref: Data Integrity]] proof is updated, since the payload for the proof's signature includes the [[ref: RevRegEntry]] the resource and the metadata (including the `links` item).
```
brianorwhatever, line 135: The `links` name isn't the greatest.. Would it be better to be more explicit? `entries`?

spec-anoncreds-method/specification.md
```diff
@@ -0,0 +1,376 @@
...
+        "type": "DataIntegrityProof",
+        "cryptosuite": "eddsa-jcs-2022",
+        "verificationMethod": "did:webvh:{SCID}:example.com#key-01"
+    }
+}
+```
+
+#### Calculating the Attested Resource `digestMultibase`
+
+Both the [[ref: Attested Resource]] creator/publisher and resolver/verifier **MUST** calculate the `digestMultibase` using: `multibase(multihash(jcs(<resource>), 'sha-256'), 'b58btc')`. The calculation **MUST** apply these specifications:
```
brianorwhatever, line 293: I see now the name hash is of only the `content`. I think the reason for this was mentioned before but I can't recall what it was. Please, educate me on why the DI proof isn't sufficient as asking verifiers to verify another custom integrity proof concerns me 😄

spec-anoncreds-method/specification.md
```diff
@@ -0,0 +1,376 @@
...
+- `multihash`: MultiHash Specification: [[spec:multihash]]
+- `multibase`: Multibase Specification: [[spec:multiformats]]
+
+#### Data Integrity Proof Generation
+
+The [[ref: Data Integrity]] proof ([[spec:VC-DATA-INTEGRITY]]) attached to [[ref: Attested Resources]] **MUST** use the `eddsa-jcs-2022` cryptosuite schema as referenced in [[spec:di-eddsa-v1.0]].
+
+#### Creation and Publishing
+
+The following process **MUST** be followed to create the an [[ref: Attested Resource]]. The input to this process is the resource -- a JSON item.
```
brianorwhatever, line 307:
```suggestion
The following process **MUST** be followed to create an [[ref: Attested Resource]]. The input to this process is the resource -- a JSON item.
```

spec-anoncreds-method/specification.md
```diff
@@ -0,0 +1,376 @@
...
+        "@protected": true,
+        "id": "@id",
+        "type": "@type",
+        "undefined": "https://www.w3.org/ns/credentials/undefined-term#",
+        "digestMultibase": {
+          "@id": "https://w3id.org/security#digestMultibase",
+          "@type": "https://w3id.org/security#multibase"
+        },
+        "mirrorLink": {
+          "@id": "undefined:mirrorLink",
```
brianorwhatever, line 351: to be defined? 😅

brianorwhatever (dismissed):
most comments are more commentary - happy to approve for merge as is and continue discussions

spec-anoncreds-method/abstract.md
```diff
@@ -0,0 +1,10 @@
+## Abstract
+
+This `did:webvh` AnonCreds Method specification defines how [[spec: AnonCreds]] objects should be registered (written) and resolved when rooting them in a `did:webvh` DID. This specification parallels other DID-specific AnonCreds methods that are registered in the [AnonCreds Methods Registry].
+
+[AnonCreds Methods Registry]: https://hyperledger.github.io/anoncreds-methods-registry/
+
+The specification introduces the concept of an [[ref: Attested Resource]] that enables verifiable identifiers that resolve to resources (files) managed by a `did:webvh` Controller. We anticipate later extracting the [[ref: Attested Resources]] concept into a separate specification on its own. Each of the published AnonCreds objects ([[ref: schema]], [[ref: CredDef]], [[ref: RevRegDef]], and [[ref: RevRegEntry]] are treated as [[ref: Attested Resources]] in this AnonCreds method.
```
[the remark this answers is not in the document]
PatStLouis: It could be defined as a resource identifier that has a verifiable component in it's uri, so once its published, part of the data is immutable (bound to the identifier).

In this case, its the last most path section being a hash of the `content` property of the file it resolves to.

spec-anoncreds-method/specification.md
```diff
@@ -0,0 +1,376 @@
...
+
+The specification of an [[ref: Attested Resource]] is found [later in this specification](#attested-resources) and may later be extracted into a standalone specification.  For those unfamiliar with [[ref: Attested Resources]], here is a brief summary of their important attributes:
+
+- [[ref: Attested Resources]] are a JSON structure that embed a complete resource (e.g., an AnonCred object, an encoded image, a [Bitstring Status List](https://www.w3.org/TR/vc-bitstring-status-list/), an [OCA Bundle](https://bcgov.github.io/aries-oca-bundles/), or any other resource published by the [[ref: DID Controller]]), plus metadata about the resource.
+  - Note that a `did:webvh` using [[ref: DID Controller]] does not have to publish resources as [[ref: Attested Resources]]. It is just a usefully verifiable data model.
+- The identifier from an Attested Resource is resolvable and contains a deterministic hash of the resource, allowing a resolver to verify that the content of the identified resource has not been altered.
+- The Attested Resource includes a [[ref: Data Integrity]] proof signed by the [[ref: DID Controller]] of the Attested Resource JSON (the resource, and metadata).
```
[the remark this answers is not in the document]
PatStLouis: The hash in name portion is a digest of the `content` property. The proof is over the whole document.

I could take your same argument and say the SCID is redundant because there is a proof on the entry, but it serves a different purpose.

If I was to publish a resource with a proof, then change the resource for another one with a proof, you wouldn't be able to tell. By binding a portion of the resource (content) to it's identifier (hash-in-name), you can't modify this content without consumers noticing.

spec-anoncreds-method/specification.md
PatStLouis: they are links. In this specific example, they represent entries, but they are abstract links. This is meant to be the same as `relatedResource` form the data model except the digest is optional.

spec-anoncreds-method/specification.md
PatStLouis: It's simply to ensure that the content doesn't change once the resource is published at a location. Data Integrity Proof wouldn't prevent that. These are mechanisms in place so a consumer can validate the integrity (DI Proof) and immutability (content hash) of the data.

spec-anoncreds-method/specification.md
```diff
@@ -0,0 +1,376 @@
...
+        "type": "@type",
+        "undefined": "https://www.w3.org/ns/credentials/undefined-term#",
+        "digestMultibase": {
+          "@id": "https://w3id.org/security#digestMultibase",
+          "@type": "https://w3id.org/security#multibase"
+        },
+        "mirrorLink": {
+          "@id": "undefined:mirrorLink",
+          "@type": "https://schema.org/URL"
+        },
```
PatStLouis, line 353:
```suggestion
```

spec-anoncreds-method/specification.md
PatStLouis: it will currently just land to [the undefined terms vocab page](https://www.w3.org/ns/credentials/undefined-term#links) until we have a vocab term to link to.

PatStLouis (dismissed):

spec-anoncreds-method/abstract.md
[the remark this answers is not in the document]
swcurran: Good point — I will update this.  Verifiable identifier — and identifier for something (in this case — an AnonCreds object) that can be verified.  In this case by verifying that the hash within the identifier matches the hash of the retrieved content, and that the proof on the object is from the expected party and verifies.

spec-anoncreds-method/specification.md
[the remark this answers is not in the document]
swcurran: I don’t think so. Like a hashlink, the hash in the name is necessary to make it part of the identifier, that the resolver gets from some 3rd party source — e.g. the holder gets from the issuer in the verifiable credential.  We want to know that the object being resolved has the same content as when the identifier was inserted into the “other object” — e.g. in the verifiable credential.  Does that make sense?

spec-anoncreds-method/specification.md
swcurran: Yes — thanks.  Will fix.

spec-anoncreds-method/specification.md
swcurran: Agreed — not my favourite.  I think we should make it more generic — enabling it to be used beyond the current use case of AnonCreds entries. “relatedData”?  Too generic?

spec-anoncreds-method/specification.md
swcurran: See my comment above.  The DI proof is across all of the data, including the resource — so the DID Controller is acknowledging the content at the time.  The metadata and related data is outside of the resource to enable use cases like AnonCreds RevRegDefs and Entries.  And the hashlink is needed for proof that nothing was altered after the identifier for the object was generated.

spec-anoncreds-method/specification.md
swcurran: @PatStLouis?  I don’t know what this is...

swcurran:
I hope @PatStLouis and I said the same things...

spec-anoncreds-method/specification.md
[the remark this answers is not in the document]
brianorwhatever: But isn't the object controlled by the issuer? ie what is the attack vector that is being prevented with this? A verifier is already verifying the proof was from the issuer. If the issuer changes the contents and create a new proof they are only hurting their own reputation / previously issued credentials. Is this solely to prevent an issuer from doing something stupid or is there a malicious angle here I am missing?

spec-anoncreds-method/specification.md
brianorwhatever: I don't know if we have a shared definition of "links". My mind goes to hyperlinks which these aren't since there is more content and hyperlinks are one way (sadly). Or do you just mean in the general sense of linking something?

spec-anoncreds-method/specification.md
brianorwhatever: no, they are mechanisms in place so a consumer MUST validate the integrity. It adds a bespoke step for verifying an `AttestedResource` - I'm not against it I'm just trying to make sure it's completely 100% necessary before requiring it. It sounds like it is but I'm just wondering the attack vector.

brianorwhatever: ahhh so is the core requirement here there is a need for an `AttestedResource` to contain both mutable and immutable data?

spec-anoncreds-method/specification.md
brianorwhatever: ah yeah I see that at the top now - seems fine for now but we should get this vocab published ASAP

spec-anoncreds-method/specification.md
```diff
@@ -0,0 +1,376 @@
...
+
+The specification of an [[ref: Attested Resource]] is found [later in this specification](#attested-resources) and may later be extracted into a standalone specification.  For those unfamiliar with [[ref: Attested Resources]], here is a brief summary of their important attributes:
+
+- [[ref: Attested Resources]] are a JSON structure that embed a complete resource (e.g., an AnonCred object, an encoded image, a [Bitstring Status List](https://www.w3.org/TR/vc-bitstring-status-list/), an [OCA Bundle](https://bcgov.github.io/aries-oca-bundles/), or any other resource published by the [[ref: DID Controller]]), plus metadata about the resource.
```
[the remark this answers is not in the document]
swcurran: I can remove, but wouldn’t it be possible for it to be an attested resource, where the BitString StatusList is the resource?  The resolver would have to know that the HTTP GET is will return an Attested Resource, not the StatusList directly...

spec-anoncreds-method/specification.md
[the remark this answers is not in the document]
PatStLouis: The original intent was to compensate the immutability feature we loose from a blockchain transaction.

This will also by design prevent 2 of the same resource from being published separately.
It will also ensure every resource has a deterministic/consistent id, instead of generating a random uuid.

While true that ultimately the issuer is responsible for generating this identifier / file content, I find a lot of benefit in having this mechanism both for the controller and consumers.

spec-anoncreds-method/specification.md
PatStLouis: we originally had:
`resourceContent`
`resourceMetadata`
`relatedLinks`

we decided to trim these properties to
`content`
`metadata`
`links`

This is meant to be an array of objects, each containing an `id` field that resolves to some data related to the attested resource. The `id` field is a URI.

spec-anoncreds-method/specification.md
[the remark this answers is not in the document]
PatStLouis: The issue is that the bitstring statuslist spec already describes a data model for the entry and the status credential, and a clear algorithm to retrieve/verify the status. It's very clear that the media type return needs to be `application/vc`, which this isn't. A bitstring statuslist credential is a `VerifiableCredential`. This would make the bitstring statuslist implementation not conformant.

spec-anoncreds-method/specification.md
brianorwhatever: > but the core content of the resource remains immutable across updates
this is the same as them just _not_ mutating the content. So adding immutability here is so that the issuer doesn't accidentally do something stupid. If the issuer is malicious I'll bet there are a number of similar attacks they'd still be able to do with this immutability in place :)

spec-anoncreds-method/specification.md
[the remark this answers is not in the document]
swcurran: Yes — I was thinking about this after. A bitstring status list is a VC, so we would have a VC as an attested resource..  And of course, since it is the same use case as our RevRegDef/RevRegEntry, you get into the same combination of immutable/mutable data.  I’ll drop it from the list.

spec-anoncreds-method/specification.md
swcurran: Leaving as links -- at least for now.

spec-anoncreds-method/specification.md
swcurran: Agreed -- where do we publish it?

spec-anoncreds-method/specification.md
swcurran: Removed.

spec-anoncreds-method/specification.md
swcurran: Fixed

andrewwhitehead (approved):

brianorwhatever (approved):

PR #180, opened 2025-02-24
Adds text and a definition about did:webvh watchers

swcurran:
Adds a definition and text about watchers, noting they are not a formal part of did:webvh because there are no defined data/interactions between DID Controllers and watchers.

We could add those -- to be discussed at the [did:webvh Work Item Meeting - 2025-02-27](https://hackmd.io/k4cIK9vQSlaeg2pdHE51IQ#Meeting---27-Feb-2025) - in a couple of ways:

- Specify a URL for each witness such that they could be contacted to get the DID Log for a given SCID.
- Add a `watchers` parameter with a list of URLs for known watchers. How a watcher becomes known is outside the scope. How a DID Controller notifies a watcher of an update is outside the scope.
- See if there is a DID Core way to do this. For example, perhaps `alsoKnownAs` could have watchers in the form `<URL>?SCID=<scid>` that either returns a the DID Log, or can be resolved like a CID.
- Other?


swcurran:
Had the wrong date in the initial comment. Correct date should have been: 

We could add those -- to be discussed at the [did:webvh Work Item Meeting - 2025-02-27](https://hackmd.io/k4cIK9vQSlaeg2pdHE51IQ#Meeting---27-Feb-2025) - in a couple of ways:

- Specify a URL for each witness such that they could be contacted to get the DID Log for a given SCID.
- Add a watchers parameter with a list of URLs for known watchers. How a watcher becomes known is outside the scope.
  - How a DID Controller notifies a watcher of an update is outside the scope.
- See if there is a DID Core way to do this. For example, perhaps `alsoKnownAs` could have watchers in the form `<URL>?SCID=<scid>` that either returns a the DID Log, or can be resolved like a CID.
- Other?

PatStLouis (approved):

Issue #131, opened 2024-11-14
Further clarification on (partial) resolution


andrewwhitehead:
In the `Read (resolve)` section there is this paragraph:

> On completing the processing and successful verification of all [entries](https://identity.foundation/trustdidweb/#term:entries) in the [DID Log](https://identity.foundation/trustdidweb/#term:did-log), respond to the DID resolution request, including the application of DID query [parameters](https://identity.foundation/trustdidweb/#term:parameters) such as `?versionId=` and `?versionTime=` with the appropriate [DIDDoc](https://identity.foundation/trustdidweb/#term:diddoc) version and content.

While I don't disagree with this there are a couple side effects to consider.

- Is it the rule that a more recent, *invalid* log entry will prevent resolution of an earlier `versionId`/`versionTime`? ie. Users would no longer be able to resolve a previous *valid* entry because the log itself appears corrupt. This seems like a good policy from a security perspective.
- I think it might be also relevant when defining the behaviour of a cache, because the entire log file would always need to be processed/verified instead of needing to account for partial verifications. We would also need to define the expiry of a corrupted log, so that an unauthenticated/unwitnessed attack on a log file could be recovered from.
- The witness verification algorithm would also depend on this decision. It becomes more complicated if parsing of the log file might abort early.

brianorwhatever:
> Is it the rule that a more recent, invalid log entry will prevent resolution of an earlier versionId/versionTime? ie. Users would no longer be able to resolve a previous valid entry because the log itself appears corrupt. This seems like a good policy from a security perspective.

I don't think this should be the behaviour as there is nothing invalid about that previous entry

> I think it might be also relevant when defining the behaviour of a cache, because the entire log file would always need to be processed/verified instead of needing to account for partial verifications. We would also need to define the expiry of a corrupted log, so that an unauthenticated/unwitnessed attack on a log file could be recovered from.

yeah definitely.. however less of an issue if we allow early resolution (previous entries aren't invalidated)

> The witness verification algorithm would also depend on this decision. It becomes more complicated if parsing of the log file might abort early.

Witnesses shouldn't be asked to sign an earlier version in fact we should disallow it. so for them processing should never abort early

swcurran:
Do we need to add any clarifications to the spec. related to these questions?  @andrewwhitehead 's document with an overview of the processing carried out by the Python implementation is helpful in defining these issues. But I don't think they change the specification.

[...]

PR #186, opened 2025-03-05
Clarifications on partial resolution and caching of DIDs that fail verification

swcurran:
Addresses issue #131.

Clarifies that if a query parameters point to a valid DID Log entry, it should be returned even if later entries are invalid.  Also talks about resolvers not caching DIDs that fail verification so that the Controller has a chance to recover an erroneous or malicious update.


brianorwhatever (approved):
Looks great

Issue #191, opened 2025-03-17
Spec doesn't specify that a Resolver MUST return an error when DID being resolved not in DIDDoc


swcurran:
From @peacekeeper 👍 

I'm doing some experiments with portability, I tried simply copying the did.jsonl file from one location to another, without actually updating it.
I used this original DID: 

`did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:gist.githubusercontent.com:brianorwhatever:9c4633d18eb644f7a47f93a802691626:raw`

And then copied the did.jsonl file to this DID: 

`did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:gist.githubusercontent.com:peacekeeper:2a0e3e9b87819a38555273a3f2c5bd2c:raw`

They are both resolvable via the universal resolver, which is based on the [did:webvh TS implementation](https://github.com/decentralized-identity/didwebvh-ts):

- [https://dev.uniresolver.io/#did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:g[…]nt.com:brianorwhatever:9c4633d18eb644f7a47f93a802691626:raw](https://dev.uniresolver.io/#did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:g[…]nt.com:brianorwhatever:9c4633d18eb644f7a47f93a802691626:raw)
- [https://dev.uniresolver.io/#did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:g[…]ontent.com:peacekeeper:2a0e3e9b87819a38555273a3f2c5bd2c:raw](https://dev.uniresolver.io/#did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:g[…]ontent.com:peacekeeper:2a0e3e9b87819a38555273a3f2c5bd2c:raw)
- 
But since I didn't actually update the DID document, resolving the "new" DID actually returns exactly the same DID document as the original DID. In other words, the value of the "id" property of the DID document does not match the DID that was resolved, which I think should be an error.

swcurran:
[...]

PR #192, opened 2025-03-19
Clarify that the DID being resolved must be the DIDDoc id (Issue 191)

swcurran:
Addresses issue #191.

Notes that a DID being resolved must match the root `id` in the DIDDoc of the entry being returned or the most recent version of the DIDDoc. If not, the resolution must fail.


swcurran:
@peacekeeper -- does this address your concern expressed in #191 ?

peacekeeper (dismissed):
Looks good I think..

spec/specification.md
```diff
@@ -395,6 +395,7 @@ For each entry:
    of this specification.
 6. Get the value of the [[ref: log entry]] property `state`, which is the [[ref:
    DIDDoc]] for the version.
+      1. Track and check that the DID being resolved matches the top-level `id` in at least one version of the [[ref: DIDDoc]]; otherwise, the resolution **MUST** be flagged as invalid.
```
brianorwhatever, line 398: This makes sense I will have to update the TS implementation.

spec/specification.md
```diff
@@ -414,17 +415,18 @@ For each entry:
...
 
-10. Flag failed verifications appropriately, either invalidating the entire DID or marking all entries from the first invalid entry to the end of the log as invalid.
+10.  Flag failed verifications appropriately, either invalidating the entire DID or marking all entries from the first invalid entry to the end of the log as invalid.
 
-11. Respond to the resolution request based on verification results:
+11.  Respond to the resolution request based on verification results:
       1. If all verifications pass, resolve the DID, applying any query parameters as requested.
-      2. If the DID itself is invalid, return an appropriate error status code.
-      3. If the request includes query parameters (e.g., `?versionId=` or `?versionTime=`) that reference valid [[ref: DID log entries]], return the corresponding [[ref: DIDDoc]] version with a successful status code—even if later entries in the log are invalid.
+      2. If the request includes query parameters (e.g., `?versionId=` or `?versionTime=`) that reference valid [[ref: DID log entries]], return the corresponding [[ref: DIDDoc]] version with a successful status code—even if later entries in the log are invalid.
```
brianorwhatever, line 428: So is the proper thing to do so is the proper thing to do here, verify the entire log and then go back and get the proper version ID resolving metadata that is partially from the latest log entry and partially from the resolved version? TS currently returns early when resolving an old version

brianorwhatever (approved):

spec/specification.md
swcurran: Ah…yes.  Some of the metadata is from the version resolved, some from the latest valid version.  The idea is that it is not useful to provide the `deactivate` or list of watchers from the resolved version if they are updated later.

swcurran: This is covered in the spec in a comment just above this section (it was merged in a separate PR).  If you request a particular version that is valid it will be returned, even if a later version is invalid.   When that happens, the last valid version is defined as the last version, and the invalid version(s) after are ignored.

brianorwhatever: ah ok interesting will have to add a bit more logic for that

PatStLouis (approved):

Issue #43, opened 2024-04-11
Additional resolver metadata


andrewwhitehead:
It may be useful, especially for implementing the DNS extensions, for clients to receive information on the associated document proofs. The DID resolution metadata could include this information:

- proof(s) for the resolved version
- associated raw (multibase?) public keys for each referenced verification method
- possibly the hash of the associated log line

swcurran:
Looking at the Universal Resolver, the following is the current DID Resolution Metadata returned by the `did:webvh` resolver ([example](  https://dev.uniresolver.io/#did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:gist.githubusercontent.com:brianorwhatever:9c4633d18eb644f7a47f93a802691626:raw))

```json
{
  "driverDuration": 188,
  "contentType": "application/did+ld+json",
  "retrieved": "2025-03-05T19:56:15.597Z",
  "pattern": "^(did:webvh:.+)$",
  "driverUrl": "http://driver-did-webvh:8080/1.0/identifiers/",
  "duration": 189,
  "did": {
    "didString": "did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:gist.githubusercontent.com:brianorwhatever:9c4633d18eb644f7a47f93a802691626:raw",
    "methodSpecificId": "QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:gist.githubusercontent.com:brianorwhatever:9c4633d18eb644f7a47f93a802691626:raw",
    "method": "webvh"
  },
  "didUrl": {
    "path": null,
    "fragment": null,
    "query": null,
    "didUrlString": "did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:gist.githubusercontent.com:brianorwhatever:9c4633d18eb644f7a47f93a802691626:raw",
    "parameters": null,
    "did": {
      "didString": "did:webvh:QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p:gist.githubusercontent.com:brianorwhatever:9c4633d18eb644f7a47f93a802691626:raw",
[...]

Issue #23, opened 2024-03-29
List of resolver verification errors and error codes


swcurran:
I thought it would be useful to add a ticket where we can track a list of error conditions in the resolver that are possible and what error codes (or other technique) we can use to surface the error to the caller.

* Log not found.
* etc.

swcurran:
Need help from implementers to get a list that we can then enumerate into Codes and Text.

swcurran:
@brianorwhatever @andrewwhitehead -- can you extract a list of error codes that get returned from your implementations.  Assuming we agree that they should go into the spec.

swcurran:
@peacekeeper -- looking at the DID Resolution Metadata in the DID Resolution spec. and DID Resolution Extensions repos.  There is really not much there -- just a reference to an `error` item and a handful of error codes.  From what I can tell, the only two errors applicable to did:webvh are `notFound` and `invalidDid` , where the latter is the catch all for any DID Resolution error that occurs.

Is there a convention (or standard) that defines a string that can be added beside the error item where the resolver could put additional information about the cause of the error?  For example, I'm proposing adding this to the did:webvh spec -- but if others have done this before, I'd like to be consistent:

> A resolver **SHOULD** include a `reason` item in the DID Resolution Metadata that provides additional information about why the error occurred. A non-normative document on the [did:webvh Information Site](https://didwebvh.info) provides a list of reasons defined by resolver implementers to aid in debugging DID resolution errors.

What do you think?  Any other guidance on how to document the resolution errors?  Any other DID Methods that cover this that I should look at and copy?

swcurran:
Resolved in #193

PR #193, opened 2025-03-26
Defines the DIDDoc Metadata and DID Resolution Metadata errors to be returned by a did:webvh resolver

swcurran:
Adds DIDDoc Metadata and DID Resolution Metadata errors to the specification, addressing issues: #43 and #23.

Error code data is based on feedback from @brianorwhatever from the [didwebvh-ts](https://github.com/decentralized-identity/didwebvh-ts) implementation. I added a few error codes (catch-all error, DID URL not found) and adjusted some of the error messages/triggers. The error messages are TS specific -- templates with placeholders for data. It would be good if we kept those. There are no error codes provided for these, no indicator that resolvers may want to provide other errors, and no "success" indicator (assuming that is in the DID Resolution spec.

The DID Document Metadata is a list of the parameters that a client might want to know about a did:webvh DID. Deliberately left off are parameters that matter only to a DID resolver and require the full DID Log to use.

To be discussed at the March 27, 2025 Work Item meeting -- [Agenda and Call In Info](https://hackmd.io/k4cIK9vQSlaeg2pdHE51IQ).


spec/specification.md
```diff
@@ -428,15 +428,57 @@ For each entry:
...
+  "versionId": "1-QmRRaLXwc6BjBuBPosSupJwEQ8w9f3znP7yfbpGfwcnLr6",
+  "versionTime": "2025-01-23T04:12:36Z",
+  "created": "2025-01-23T04:12:36Z",
+  "updated": "2025-01-23T04:12:36Z",
+  "scid": "QmPEQVM1JPTyrvEgBcDXwjK4TeyLGSX1PxjgyeAisdWM1p",
+  "portable": false,
+  "deactivated": false,
+  "witness": { ...
+  },
+  "watchers: [ ...
```
brianorwhatever, line 444: I'm confused on why this is the top level property but in the other PR what URLs are part of the witness object. Will make some comments in the PR as well

brianorwhatever: Yeah, I think including witnesses makes sense in the resolution what I'm questioning is more of a question for the watchers PR.

brianorwhatever (approved):

spec/specification.md
swcurran: That is absolutely required.  A resolver client that is unsure that a DID will exist for as long as it needs it would want to keep a list of places where it can be retrieved.  I think that is very important.

PatStLouis: It would be nice to have the latest witness rules in the metadata.

PatStLouis (approved):

swcurran:
@PatStLouis — what does your last comment mean?  The witness data is in the metadata — are you saying that is nice, or are you saying there is something else that should be included?

PatStLouis:
@swcurran yes I'm saying its useful to have it there. I might of misunderstood the comment exchange on this PR as potentially wanting to remove it. The way its defined now is good IMO. I approved this PR

PatStLouis (approved):

PR #196, opened 2025-04-03
Extend Security and Privacy Considerations section

AlexanderShenshin:
- Added new `did:webvh`-specific considerations
- Updated `In-transit Security` point to mention `watchers` and `witnesses`
- Misc. cleanup

My thoughts and potential points to discuss

- General approach for "Security and Privacy" - non-normative with `did:webvh`-specific points seems to be good enough
  - Other option would be to describe privacy/security properties based on RFCs (for example, similar to how it's done in [DID Indy Method spec](https://hyperledger.github.io/indy-did-method/#privacy-considerations)). However, `did:webvh` do not describe or enforce all related implementation options, so sticking to RFC-based list of properties may not provide much value
- `did:webvh` spec mentions some interesting use cases that, however, not part of the spec. Good example is [watchers network to independently monitor DID updates via consensus](https://github.com/decentralized-identity/didwebvh/blob/main/spec/specification.md?plain=1#L1036), being also a good candidate for providing additional guidance as "Security Consideration". But, as these are out of scope for the spec, we should leave them out of "Security and Privacy" as well

### DIFF
--- a/spec/security_and_privacy.md
+++ b/spec/security_and_privacy.md
@@ -4,7 +4,7 @@
 
 #### DNS Security Considerations
 
-Implementers must secure DNS resolution to protect against attacks like Man in
+Implementers **MUST** secure DNS resolution to protect against attacks like Man in
 the Middle, following the detailed guidance in the [did:web
 specification](https://w3c-ccg.github.io/did-method-web/#dns-considerations).
 The use of DNSSEC [[spec:RFC4033]], [[spec:RFC4034]], [[spec:RFC4035]] is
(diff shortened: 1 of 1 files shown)


swcurran (approved):

swcurran:
Good stuff — thanks @AlexanderShenshin .  You might want to add your name to the authors.

AlexanderShenshin:
@swcurran Thank you!
I've created a small PR for spec header, added myself as editor + put a reference to Rust implementation: https://github.com/decentralized-identity/didwebvh/pull/197

Issue #203, opened 2025-04-24
Create a JSON Schema definition for a DID Log Entry


swcurran:
Define a JSON Schema definition for a DID Log Entry, including all of the possible `parameter` attributes.  Include a link in the spec to the JSON Schema.

swcurran:
@PatStLouis -- is this a good one for you?  I think you've done a fair amount of JSON Schema work previously, right?

swcurran:
Resolved by #206. Leaving open until the JSON Schema is referenced in the spec.

PR #206, opened 2025-04-29
JSON Schema for version 1.0 log entry

PatStLouis:
Addresses #203


swcurran:
@andrewwhitehead / @brianorwhatever — could one of you please review this and confirm it?

How is something like this typically published?  Do we use the raw GitHub URL?  Find a way to have it published with the spec?

PatStLouis:
it could be just published as part of the gh page

swcurran:
Doesn’t need to be a standalone file that can be pulled in?

PatStLouis:
I just changed the file location so it lives in its respective spec version directory. Its better I think?

PatStLouis:
> Doesn’t need to be a standalone file that can be pulled in?

I'm not sure I understand what you're asking

swcurran:
No — I don’t think so. I think it should be in a separate folder, named by version like was done for the OpenAPI file.  It is a file that is published and referenced in the spec, but needs to be a in stable place, vs. copied around with the rest of the spec.  Does that make sense?

swcurran:
> > Doesn’t need to be a standalone file that can be pulled in?
> 
> I'm not sure I understand what you're asking

I assumed that JSON tooling (such as in IDEs or CI/CD pipelines) would need to access the file via its URL.  Is that not right?  We would not want to embed it in a GH-Pages file.

If you mean publishing it as a separate file as part of the GH-Pages — yes, but I don’t know how to do that with Spec-Up :-). Do you?

swcurran:
> it could be just published as part of the gh page. Like the Watcher OpenAPI: https://identity.foundation/didwebvh/watcherOpenAPI/watcher-v1.0.0.yml
> 
> This schema would be available under: https://identity.foundation/didwebvh/schemas/v1.0/schema.json

DOH!  I missed this comment (or at least checking the links) -- hence my confusing reply.  I didn't realize that the OpenAPI file was published automagically on gh-pages like that.

I suggest moving the file back from the spec folder and to where you had it and have it autopublish from there.

andrewwhitehead:
I think possibly the `versionTime`, `created` and `expires` attributes could use the JSON Schema `date-time` type?

I see `updateKeys` is marked required for the parameters, but that is not generally the case (it does not have to be updated every time). We could potentially have a restricted schema for the first line of the log? Similarly, I don't think the `proof` property is strictly required if a log entry has not been signed yet, although I suppose we could require an empty list.

PatStLouis:
moved the schema back in the previous location, added the date-time type and removed some of the restrictions

swcurran (approved):

Issue #208, opened 2025-05-02
Links in the v1.0 spec landing page version are broken links


swcurran:
Per the (excellent) request  #199 to have a permalink for the current version (e.g. <URL>/v1.0) and have the main landing page also contain the current spec (e.g. <URL>).

I tried to do that by just rendering the current version spec twice, once at the root, and once at the /v1.0 location -- via two entries in the specs.json file.  But of course, that breaks the relative links to the other spec versions (e.g. from the root, I need "v.05" and from /v1.0 I need "../v0.5". Doh!

Fixing that with a new approach.

PR #209, opened 2025-05-02
Fix the broken links on the landing page, add way to control the current default version of the spec

swcurran:
Resolves issue #208.

Provides a way (documented in the Readme) for us to control the "default" version of the spec to appear when hitting the landing page ([https://identity.foundation/didwebvh/](https://identity.foundation/didwebvh/)) by editing a simple `index.html` file in the root folder that is a redirect.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 package-lock.json
 index.html
+!/index.html
 fonts
 
 # Logs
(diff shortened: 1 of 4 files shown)


brianorwhatever (approved):
I think this looks good. My only concern is with the relative path, if that's going to work with the didwebvh path. only one way to find out though let's merge!

also maybe worth noting SEO might not be the best with this but I think it's probably fine these days who knows what's happening on search engines..

swcurran:
Nice!  Works well.  Not sure about SEO, but all of our mkdocs sites (e.g. https://didwebvh.info) work that way.

It’s also a nicer solution to controlling what is the current “default” version.

PR #231, opened 2025-07-07
Update the guidance for publishing a parallel did:web DID to include services

### LINKED ISSUE #223: Add guidance on creating a parallel did:web for did:webvh services and alsoKnownAs
Extend the guidance in the section on creating a parallel did:web to a did:webvh to include:

- Add the implicit did:webvh services that get added to each did:webvh if not specified.
- Add to the did:webvh DID Doc the `alsoKnownAs` for the did:web.

swcurran:
Resolves [Issue #233](https://github.com/decentralized-identity/didwebvh/issues/223) by adding requirement to add the implicit services (if any) from did:webvh to did:web and making the `alsoKnownAs` reciprocal -- that it should also be in the `did:webvh` DIDDoc.


PatStLouis (commented):
I definitely approve the bi directional `alsoKnownAs` recommendation. I supposed it also makes sense to use the services although it could create ambiguity about who signs the whois.vp if the service is included in both locations. Maybe it should be signed with the webvh version or maybe it doesn't matter (could be signed by one or another, if there is bidirectional `alsoKnownAs` reference only). This should likely be also addressed.

PatStLouis:
I mean adding a note about this in the whois.vp section of the spec, that in the case of a parallel did, this presentation could be signed by one or the other, or both.

swcurran:
Good enough @PatStLouis ?

PatStLouis (approved):

PR #233, opened 2025-07-09
Clarify the general handling of parameters, including defaults and deactivations

swcurran:
This PR is mostly updates to clarify some general rules around parameters, and to be precise about the handling of each parameter. However, there are changes that break the previous descriptions, as listed below.

- Changes the recommended deactivation step for updateKeys from null to []
- Changes the default for nextKeyHashes from null to [].
- Changes the action to deactivate pre-rotation to setting nextKeyHashes to[]instead of null.
- Changes the default for witness from null to `{}`.
- Changes the action to deactivate witnessing to setting witness to {} instead of null.
- Changes the default for watchers from null to [].
- Adds the action to deactivate watchers from setting watchers to[].
- Sets the previously undefined default to 3600 (1 hr)

Please review this from three perspectives:

- The general accuracy of the new section -- is it good? Helpful?
- Should the changes be made that break the previous definition?
- If we accept the changes, do we have to update the version?

The last two give us three paths forward -- all of which are open to us.

1. Keep the clarifications, but eliminate the changes from the  "old" ways.
2.  Accept the new ways and don't update the version.
3. Accept the new ways and update the version.

As currently written, the PR takes the second path -- no change to the version, but an update to the defaults.

Particulary important for implementers!


swcurran:
Feedback welcome -- @brianorwhatever @andrewwhitehead @stormer78 @PatStLouis @AlexanderShenshin @SylvainMartel

stormer78:
Would drop the null usage as well. It causes a lot of confusion and hard to handle in true "typed" languages

Would keep version to 1.0 for now, dealing with multiple implementation versions so early would be more problematic

TTL is good at an hour as a sensible default.

AlexanderShenshin:
Changes look good to me and there are good points on TTL raised by @stormer78

Regarding version change approach - I don't have a strong opinion here, but I think it would be more appropriate to change the version due to breaking previous rules/descriptions.

swcurran:
> Should `ttl` be added to the metadata? Clients of the resolver may also be implementing basic caching, and it would be good to easily pass the TTL down to them.
> 
> Would default to 3600 unless a value has been provided.

Yes -- it definitely should be in the metadata.  Thanks for noting that.  A client should have visibility on that.

spec/specification.md
```diff
@@ -625,136 +625,88 @@ of those processes is specified in the following sections.
 
 #### `did:webvh` DID Method Parameters
 
-Entries in the `did:webvh` [[ref: DID Log]] contain the JSON object `parameters`
-that define the DID processing [[ref: parameters]] being used by the [[ref: DID
-Controller]] when publishing the current and subsequent [[ref: DID log
-entries]]. A DID Resolver **MUST** use the same [[ref: parameters]] when
-processing the [[ref: DID Log]] to resolve the DID. The `parameters` object
-**MUST** only include properties defined in this specification.
+All `did:webvh` [[ref: Log entries]] contain the JSON object `parameters`. This object defines the DID processing [[ref: parameters]] used by the [[ref: DID Controller]] when publishing the current and subsequent [[ref: DID log entries]]. DID Resolvers **MUST** use the same [[ref: parameters]] to process the [[ref: DID Log]] to resolve the DID. The `parameters` object **MUST** only include properties defined in the version of the `did:wenbvh` DID Method specification being used.
```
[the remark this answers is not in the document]
swcurran: Fixed

spec-v1.0/specification.md
```diff
@@ -625,144 +627,88 @@ of those processes is specified in the following sections.
...
+- `scid`: The [[ref: SCID]] value for the DID.
+  - **MUST** appear in the first [[ref: log entry]].
+  - **MUST NOT** appear in later [[ref: log entries]].
+- `updateKeys`: A JSON array of [[ref: multikey]] formatted public keys associated with the private keys that are authorized to sign the log entries that update the DID. See the [Authorized Keys](#authorized-keys) section of this specification for additional details.
+  - This property **MUST** appear in the first [[ref: log entry]] and **MAY** appear in subsequent entries.
+  - If not present in later [[ref: DID log entries]], the previous value continues to apply.
+  - A key from the active `updateKeys` array **MUST** be used to authorize the each [[ref: log entry]], where active is defined as follows.
+    - In the first [[ref: log entry]], the active `updateKeys` is the one defined in that entry.
+    - In all other [[ref: log entries]] *without* [[ref: Key Pre-Rotation]] active, the active `updateKeys` is that of the most recent **prior** [[ref: log entry]].
+    - In all other [[ref: log entries]] *without* [[ref: Key Pre-Rotation]] active, the active `updateKeys` is that of the most current [[ref: log entry]].
```
PatStLouis, line 679: Is there a difference between these 2 statements? Isn't the second meant to apply if the Pre-Rotation is active?

spec-v1.0/specification.md
```diff
@@ -625,144 +627,88 @@ of those processes is specified in the following sections.
...
+- `deactivated`: A JSON boolean that indicates whether the DID has been deactivated. A deactivated DID is no longer subject to updates but remains resolvable. See the [deactivate (revoke)](#deactivate-revoke) section of this specification for more details.
+  - Defaults to `false` if not set in the first [[ref: DID log entry]].
+  - If set to `true`, the DID is considered deactivated and no further updates to the DID are permitted.
+- `ttl`: An unsigned integer that indicates how long, in seconds, a resolver should cache the resolved `did:webvh` DID before refreshing. It provides guidance from the [[ref: DID Controller]] on cache duratio, with a range of 0 to 2^31. The parameter is analogous to the `TTL` parameter used in DNS [[spec: rfc2181]]. Caching a `did:webvh` can be valuable in places where the business rules require resolving a number of DID URLs for the same DID. For example, a client might want call the resolver to the current [[ref: DIDDoc]], and then make repeated calls to get all of the previous versions of the [[ref: DIDDoc]]. By caching the [[ref: DIDDoc]] state, the resolver would not have to retrieve and process the [[ref: DID Log]] on each call.
```
PatStLouis, line 709:
```suggestion
- `ttl`: An unsigned integer that indicates how long, in seconds, a resolver should cache the resolved `did:webvh` DID before refreshing. It provides guidance from the [[ref: DID Controller]] on cache duration, with a range of 0 to 2^31. The parameter is analogous to the `TTL` parameter used in DNS [[spec: rfc2181]]. Caching a `did:webvh` can be valuable in places where the business rules require resolving a number of DID URLs for the same DID. For example, a client might want call the resolver to the current [[ref: DIDDoc]], and then make repeated calls to get all of the previous versions of the [[ref: DIDDoc]]. By caching the [[ref: DIDDoc]] state, the resolver would not have to retrieve and process the [[ref: DID Log]] on each call.
```

spec/specification.md
```diff
@@ -625,144 +627,88 @@ of those processes is specified in the following sections.
...
+- `scid`: The [[ref: SCID]] value for the DID.
+  - **MUST** appear in the first [[ref: log entry]].
+  - **MUST NOT** appear in later [[ref: log entries]].
+- `updateKeys`: A JSON array of [[ref: multikey]] formatted public keys associated with the private keys that are authorized to sign the log entries that update the DID. See the [Authorized Keys](#authorized-keys) section of this specification for additional details.
+  - This property **MUST** appear in the first [[ref: log entry]] and **MAY** appear in subsequent entries.
+  - If not present in later [[ref: DID log entries]], the previous value continues to apply.
+  - A key from the active `updateKeys` array **MUST** be used to authorize the each [[ref: log entry]], where active is defined as follows.
+    - In the first [[ref: log entry]], the active `updateKeys` is the one defined in that entry.
+    - In all other [[ref: log entries]] *without* [[ref: Key Pre-Rotation]] active, the active `updateKeys` is that of the most recent **prior** [[ref: log entry]].
+    - In all other [[ref: log entries]] *without* [[ref: Key Pre-Rotation]] active, the active `updateKeys` is that of the most current [[ref: log entry]].
```
PatStLouis, line 679: Same as above

spec/specification.md
```diff
@@ -625,144 +627,88 @@ of those processes is specified in the following sections.
...
+- `deactivated`: A JSON boolean that indicates whether the DID has been deactivated. A deactivated DID is no longer subject to updates but remains resolvable. See the [deactivate (revoke)](#deactivate-revoke) section of this specification for more details.
+  - Defaults to `false` if not set in the first [[ref: DID log entry]].
+  - If set to `true`, the DID is considered deactivated and no further updates to the DID are permitted.
+- `ttl`: An unsigned integer that indicates how long, in seconds, a resolver should cache the resolved `did:webvh` DID before refreshing. It provides guidance from the [[ref: DID Controller]] on cache duratio, with a range of 0 to 2^31. The parameter is analogous to the `TTL` parameter used in DNS [[spec: rfc2181]]. Caching a `did:webvh` can be valuable in places where the business rules require resolving a number of DID URLs for the same DID. For example, a client might want call the resolver to the current [[ref: DIDDoc]], and then make repeated calls to get all of the previous versions of the [[ref: DIDDoc]]. By caching the [[ref: DIDDoc]] state, the resolver would not have to retrieve and process the [[ref: DID Log]] on each call.
```
PatStLouis, line 709:
```suggestion
- `ttl`: An unsigned integer that indicates how long, in seconds, a resolver should cache the resolved `did:webvh` DID before refreshing. It provides guidance from the [[ref: DID Controller]] on cache duration, with a range of 0 to 2^31. The parameter is analogous to the `TTL` parameter used in DNS [[spec: rfc2181]]. Caching a `did:webvh` can be valuable in places where the business rules require resolving a number of DID URLs for the same DID. For example, a client might want call the resolver to the current [[ref: DIDDoc]], and then make repeated calls to get all of the previous versions of the [[ref: DIDDoc]]. By caching the [[ref: DIDDoc]] state, the resolver would not have to retrieve and process the [[ref: DID Log]] on each call.
```

spec-v1.0/specification.md
swcurran: Fixed

spec/specification.md
swcurran: Fixed

swcurran:
I have added a note to the spec acknowledging that early implementations of did:webvh may not have followed the "no nulls" requirement now in the spec.  I think that is sufficient to keep the version the same, but welcome comments from others.

Please review and let me know ASAP so we can merge and move on to other clarifications.

brianorwhatever (approved):
LGTM

PatStLouis (approved):

Issue #235, opened 2025-07-16
Method can only increase


swcurran:
Make the spec specific that the `method` parameter can only increase. A DID cannot "fall back" to a previous version of the spec.

swcurran:
Add a note about not needing backwards compatibility to prior to 1.0.

PR #239, opened 2025-07-22
Clarify that the method parameter cannot decrease the specification version currently used by the DID

swcurran:
Addresses issue [\#235](https://github.com/decentralized-identity/didwebvh/issues/235) -- that the `method` parameter set in a later DID log entry must references the currently used specification version or a later one.  No degrading a DID to use a lower version specification.


PatStLouis (dismissed):

swcurran:
@brianorwhatever @PatStLouis -- last commit adds reference to [semver](https://semver.org) for the `method` string value. Should cover @brianorwhatever 's comment from this morning's meeting.

PatStLouis (approved):

Issue #241, opened 2025-07-30
Add guidance around deactivate related to the DID Resolution Spec not returning a DIDDoc after deactivation


swcurran:
In writing the deactivate part of the spec, I (we?) had thought that by deactivating the DID, a resolver would convey the deactivation in the resolution metadata (which it does) when resolving the DID to a DIDDoc. However, the DID Resolution spec says that a DIDDoc is not to be returned when the DID is deactivated.  This should be clarified in the deactivate section, and it has been suggested that a DID Controller that wants to convey that the DID is "no longer maintained", could do all the updates except setting the deactivate flag -- so that the DID is "end of life'd" (can't be updated) but it still resolvable -- e.g. still yields a a DIDDoc.

This has been requested by the Swiss e-ID team, and is consistent with how we expected it to work when deactivating.

Another proposal is that we add a query parameter to enable overriding the DID Resolution spec and return the DIDDoc of a deactivated DID.

Relevant DID Spec issues: 

* https://github.com/w3c/did/issues/687
* https://github.com/w3c/did-resolution/issues/5
* https://github.com/w3c/did-extensions/issues/622

peacekeeper:
Yes, this has been discussed in the DID WG a few times, and there was agreement that this should be supported with an extension parameter `returnDeactivatedDidDocument`. We could write a mini-spec for this parameter in DIF or elsewhere.

swcurran:
@peacekeeper -- if the resolution of a deactivated DID is for a version (by `versionId` or `versionTime`) prior to deactivation, my expectation is that the DIDDoc for that version will be returned, with the metadata including a `"deactivated": true`.  Is that correct?

Thanks!

PR #252, opened 2025-08-21
Clarify the deactivate section to cover DID Resolution spec requirements and other did:webvh options

swcurran:
Addresses [#241](https://github.com/decentralized-identity/didwebvh/issues/241) about deactivation, providing both the required behaviour to align with the DID Core/DID Resolution specs, and guidance on signalling a DID is not maintained without deactivating the DID. Covers that a DIDDoc is not returned for a deactivated DID (per DID Resolution), and options for having the DID continue to be returned.


stormer78:
Read the changes - clear to understand.

The DID spec itself makes this a little clumsy as you call out, the final LogEntry will not be returned when deactivated, but any earlier versions will be returned even though the metadata says deactivated - which means you have the following behaviour:

1. If I try to resolve the latest (unspecified) DID, I will get an error and no DID Doc (how do I find out the latest version)?
2. If I try to resolve a specific version, I will get the DID Doc and metadata letting me know it is deactivated
3. If I want to find the latest version, I would need to do two queries
  a: Lookup unspecified DID, get the NO DID-DOC but metadata will tell me the latest verson
  b: Do a second lookup of the latest versionId to get the last DID Doc
  
 Question would be what is the last valid DID Doc?
 1. State of the final LogEntry?
 2. State of the LogEntry prior to the deactivated record?
 
 This probably should be clarified

swcurran:
My opinion.  The last valid DIDDoc value is the value set in the last log entry.  Why?  

- The DID Resolution spec has proposed that you can use a parameter `returnDeactivatedDidDocument` to return the DIDDoc after deactivation.  
- I think you can specify the versionId as a parameter to get back the last version.
- The DID Controller has expressed the last `state` as the value of the DIDDoc.
- Anyone can bypass a resolver, retrieve the Log and process it to get the state set by the DID Controller on the last update.

@peacekeeper -- any thoughts on this?

stormer78:
@swcurran In section 3.6.2, should it be clearer that the "deactivated" attribute in the metadata must be set to TRUE for any prior version resolution if the DID has been deactivated in later log_entries?

It somewhat implies this, but stating it more simply will make it easy for people to understand that once a DID has been deactivated, all subsequent resolutions, regardless of query, will always include deactivated = true in the metadata

spec-v1.0/specification.md
```diff
@@ -601,24 +601,15 @@ section of this specification.
...
+
+To deactivate a `did:webvh` DID per the [[ref: DID-CORE]] specification, the [[ref: DID Controller]] **MUST** add to the [[ref: DID log entry]] [[ref: parameters]] the property name and value `"deactivated": true`. Once done, a resolver **MUST NOT** return the [[ref: DIDDoc]] and **MUST** include `"deactivated": true` in the DID Resolution Metadata. A [[ref: DID Controller]] deactivating a `did:webvh` DID **MAY** update the [[ref: DIDDoc]] (e.g., setting it to {}) and some [[ref: parameters]] attributes to further indicate the deactivation of the DID, such as setting the `updateKeys` array to `[]`, preventing further versions of the DID. If the DID is using [[ref: pre-rotation]], two [[ref: DID log entries]] are required to accomplish that state: the first to stop the use of pre-rotation, and the second to set `updateKeys` to []. For additional details about turning off [[ref: pre-rotation]] see the [pre-rotation](#pre-rotation) section of this specification.
```
brianorwhatever, line 606: `{}` isn't a valid DID Doc it at least needs an `id` parameter still. Not sure if this matters in this case though

brianorwhatever (approved):
Looks good

PR #290, opened 2026-07-21
chore(abnf): explicit ABNF clarification

rmlearney-digicatapult:
Tighten `did:webvh` identifier and `DID-to-HTTPS parsing rules` to remove ambiguity and reference `DID Core`.

* Method-Specific Identifier section now explicitly layers `did:webvh` constraints on top of `DID Core ABNF`, with clearer productions for `SCID`, `domain`, optional `percent-encoded port`, and `path` segments
* Deterministic parsing requirements for `percent-encoded port` handling, including explicit disambiguation of terminal percent-encoded colon + digits and constraints on where percent-encoded colons may appear
* Explicit normalization and validation requirements for domain and path processing (single decode pass, IDNA handling, IP-address rejection, path safety checks, canonical encoding expectations)
* Clarified that DID URL `path`/`query`/`fragment` components are outside `method-specific-id` parsing and must not be interpreted as `SCID`/`domain`/`port`/`path` segments

These changes are consistent with `did:web` as well (except for the existence of `SCID`), just spelled out more explicitly.


swcurran (approved):
LGTM

PR #296, opened 2026-09-18
Cleanup of refs and defs in v1.0 spec

swcurran:
A long overdue cleanup pass of the v1.0 and "next" versions of the spec that mostly focuses on fixing defs and refs to make them in sync. The two versions "/spec" (Editor's Draft) and "/spec-v1.0" (v1.0) remain identical except for the header. This is (likely) the last update to the v1.0 spec, and the Editor's Draft will now start to move forward past v1.0.

There are only two actual text changes in this:

- The section headers in the DID URL section were screwed up, with a hanging, unused `/WHOIS" header. Those were cleaned up, with the links targeting those sections also cleared up.
- There was a link that names the old "Privacy and Security considerations" section that hadn't been changed when that section was split into two. Not really needed, but ¯\\\_(ツ)_/¯

The rest of the changes were laregly identified by Claude Code passes and follow up reviews. Here is the list of those changes:

- 21 line-wrapped [[ref: ...]] tags rejoined onto one line — these were silently deleting words like "Controller", "Log", "Entry", "Presentation", "Credentials" from the rendered spec with no error. No wording was changed, just restored.
- [[ref: DID-CORE]] → [[spec:DID-CORE]] in the deactivation section, matching how it's cited everywhere else.
- [[ref: key]] and [[ref: Update Key]] de-linked to plain text since no such terms are defined.
- Added Pre-Rotation Key as an alias on the existing Pre-Rotation definition so that reference resolves.
- [[witnesses]] → [[ref: witnesses]] in overview.md — this one was rendering as literal [[witnesses]] text in the published spec since it was missing its type prefix.
- [[ref; parameter]] → [[ref: parameter]] in the ttl parameter description — a semicolon typo instead of a colon, same literal-markup-leak bug as #5, in a different spot.
- 8 broken internal anchor links fixed (5 distinct targets), mostly caused by the heading restructuring done in this cleanup: #whois-linkedvp-service → #did-url-whois-linkedvp-service (4 occurrences), #security-and-privacy-considerations → #security-considerations, #did-url-path-resolution-service → #did-url-path-resolution, #watchers → #did-watchers, #did-to-https-transformation → #the-did-to-https-transformation.
- One dead external link fixed: https://didwebvh.info/latest/implementers_guide/#witnesses (verified 404) → https://didwebvh.info/latest/implementers-guide/witnesses/ (verified live).
- Typos: "implementors" → "implementers"; duplicated word "can be be" → "can be" (2 occurrences).
- Minor formatting cleanup: removed 2 stray trailing spaces, added a missing blank line before a ::: container closing a fenced code block, normalized 10./11. ordered-list marker spacing to match the rest of the list.


brianorwhatever (dismissed):
Nice cleanup. I think there are still some multiline [[ref: ...]] references that should be included in this pass, especially given the rendering issue described in the PR. For example, spec/specification.md still has multiline refs for parameters, witnesses, Verifiable Presentation, ISO8601, SCID, etc. There are also a couple remaining in definitions.md and overview.md, with the same occurrences in spec-v1.0/.

It would be good to catch all remaining [[ref: ...]] tags containing newlines before merging so we don’t leave any of the silent rendering failures behind.

swcurran:
Thanks, @brianorwhatever -- I trusted Claude to catch them all. Doh...  This time I used a regex -- and then almost missed updating the v1.0 spec in sync with the editor's draft.  Doh again...

I also fixed a couple of external spec references that were missing, notably the one for DID-RESOLUTION.  That's not a good one to miss.  Added to the specs.json file.

Building on this PR, so hoping to get it merged soon.

brianorwhatever (approved):
Looks good

PR #297, opened 2026-09-21
Update Editors Draft to align DID URL Path Handling with DID Resolution spec

swcurran:
Updates all references in the spec about "DID URL resolution" to "DID URL dereferencing" in alignment with the DID Resolution spec, and updates the implicit services to use what I hope will be the final approach to DID URL Path handling -- including the PathHandler and PathService DID service `type` values.  That last part is not yet in the DID Resolution spec, but I'm ever hopeful.

Includes adjustments to the READ DID Method operation (mostly just clean up in the handling of the implicit services), and a reorganization of the DID URL Path handling section to remove redundency.

~~I'll update the gh-pages publishing of my fork of the spec to render this branch so users can see the rendered changes and compared to the current text. I'll add URLs for both versions in a comment after this PR is submitted.~~

NOTE: This is the first update to the Editor's Draft that isn't in sync with v1.0. From now on, the Editor's Draft will be evolved towards the next version (1.1 or perhaps 2.0) of the did:webvh DID Method specification. The [DID v1.0](https://identity.foundation/didwebvh/) remains the current version of the spec.

- **Progress on cleaning up DID URL Handling**
- **Update Editors Draft to align DID URL Path Handling with DID Resolution spec**


swcurran:
I'm unable to publish the branch from my fork for review, so that is not available.

I also so that a fix to the v1.0 specification did sneak into this PR -- sorry about that. It is a reasonable fix, so not the end of the world, but I didn't intend it. Not sure how to remove it from the PR, so leaving it for now.  If we want it removed let me know and I will try harder.

rmlearney-digicatapult (dismissed):
👍 

Can you back-port the small change to the v1.0 spec as a separate PR?

swcurran:
Removed the change to the v1.0 spec.  Will add a separate PR for that -- and one other v1.0 spec adjustment.

rmlearney-digicatapult:
Should the editors draft not be branched from the final 1.0 version incorporating the body changes (minus header) of #298?

You're missing the references to `LINKED-VP` and `DID-EXTENSION-RESOLUTION` in this PR for example

